PR #9, opened 2019-06-29
tickets/DM-20129

fred3m:
Add unit tests to ensure the package is working and to track changes in a simple blend.


python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -15,26 +15,31 @@ class LsstSource(ExtendedSource):
...
+        if not pointSource:
+            try:
+                super().__init__(frame, center, observation, bgRms, thresh,
+                                 symmetric, monotonic, centerStep, **componentKwargs)
+                initialized = True
+            except SourceInitError:
+                # If the source is too faint for background detection,
+                # initialize it as a PointSource
+                pass
+        if not initialized:
```
natelust, line 34: Should this happen if the pointSource is set to False?

natelust: I see the above comment, but am just wondering the logic, should there not just be an error thrown or something?

tests/test_deblend.py
```diff
@@ -0,0 +1,118 @@
...
+from lsst.meas.extensions.scarlet import ScarletDeblendTask
+from lsst.afw.table import SourceCatalog
+from lsst.afw.detection import MultibandFootprint
+from lsst.afw.image import Image, MultibandImage
+
+from utils import initData
+
+
+class TestDeblend(lsst.utils.tests.TestCase):
+    def test_task(self):
```
natelust: Maybe use a more descriptive test name, as these will show up in various reports

tests/utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def numpyToStack(images, center, offset):
+    """Convert numpy and python objects to stack objects
+    """
+    cy, cx = center
+    bands, height, width = images.shape
+    x0, y0 = offset
+    bbox = Box2I(Point2I(x0, y0), Extent2I(width, height))
+    spans = [Span(y, x0, x0+width-1) for y in range(y0, y0+height)]
```
natelust: You can just make a spanset from a box here

tests/utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # Convolve the image with the psf in each channel
+        # Use scipy.signal.convolve without using FFTs as a sanity check
+        images = np.array([scipy.signal.convolve(img, psf, method="direct", mode="same")
+                           for img, psf in zip(images, psfImages)])
+        # Convolve the true morphology with the target PSF,
+        # also using scipy.signal.convolve as a sanity check
+        morphs = np.array([scipy.signal.convolve(m, targetPsfImage, method="direct", mode="same")
+                           for m in morphs])
+        morphs /= morphs.max()
```
natelust, line 86: This may just be me reviewing quickly without thinking through too much, but do you want the max or the sum in the normalization

tests/utils.py
```diff
@@ -0,0 +1,91 @@
...
+        psfImages /= psfImages.sum(axis=(1, 2))[:, None, None]
+
+    channels = range(len(images))
+    return targetPsfImage, psfImages, images, channels, seds, morphs, targetPsf, psfs
```
natelust, line 90: When you have thing many parameters consider returning a pipe_base.Struct, collections.namedtuple, or types.SimpleNamespace so that it is easier to reason and work with so many objects.

natelust (approved):
Few question/comments mostly looks fine.

python/lsst/meas/extensions/scarlet/source.py
fred3m: Part of this might be the nomenclature. Both `ExtendedSource` and `PointSource` are used to model extended sources, they just differ in their initialization. An `ExtendedSource` uses a multiband coadd to attempt to initialize large extended sources but requires a certain signal to noise level to function properly. Very faint sources barely above the background will not have enough flux to meet this requirement, so we initialize them as a point source but they are still allowed to grow.

tests/utils.py
fred3m: The max. This seems to be a more stable normalization, most likely because the center of most objects is better behaved. Recall that the value of a pixel in the model is the product of its SED and the morphology in that pixel. Since the center of an object has a higher S/N it is easier for the optimization algorithm to correctly fit the center and takes more time to fit the outside of a source. If we instead use a normalization of the entire morphology then fiddling with flux on the outskirts of an object also adjusts the flux in the center (due to the normalization of *all* the pixels) and takes longer to converge.

tests/utils.py
fred3m: I know what you mean but I'm not a big fan of returning a new class just for a return value.

tests/utils.py
fred3m: I didn't realize that you could initialize with `SpanSet(bbox)` until I actually looked at the code. It might be a good idea to add this initialization to the docs.

fred3m: Created DM-20776

PR #13, opened 2019-10-17
tickets/DM-21808: Fix PSF initialization bug and create blank source

fred3m:


fred3m:
These changes were included in DM-21965.

PR #16, opened 2020-01-13
tickets/DM-22137

fred3m:
Integrate scarlet 1.0 into the stack and test on HSC data.

### DIFF
--- a/python/lsst/meas/extensions/scarlet/__init__.py
+++ b/python/lsst/meas/extensions/scarlet/__init__.py
@@ -25,3 +25,4 @@ from .observation import *
 from .source import *
 from .blend import *
 from . import deblend
+from .version import *
--- a/python/lsst/meas/extensions/scarlet/__init__.py
+++ b/python/lsst/meas/extensions/scarlet/__init__.py
@@ -24,4 +24,4 @@
 from .observation import *
 from .source import *
 from .blend import *
-from .deblend import *
+from . import deblend
--- a/python/lsst/meas/extensions/scarlet/deblend.py
+++ b/python/lsst/meas/extensions/scarlet/deblend.py
@@ -177,7 +177,7 @@ class ScarletDeblendConfig(pexConfig.Config):
     - Other: Parameters that don't fit into the above categories
     """
     # Stopping Criteria
-    maxIter = pexConfig.Field(dtype=int, default=100,
+    maxIter = pexConfig.Field(dtype=int, default=300,
                               doc=("Maximum number of iterations to deblend a single parent"))
     relativeError = pexConfig.Field(dtype=float, default=1e-3,
                                     doc=("Change in the norm of each parameter between"
--- a/python/lsst/meas/extensions/scarlet/deblend.py
+++ b/python/lsst/meas/extensions/scarlet/deblend.py
@@ -355,16 +355,17 @@ class ScarletDeblendTask(pipeBase.Task):
         self.morphNotConvergedKey = schema.addField('deblend_morphConvergenceFailed', type='Flag',
                                                     doc='scarlet morph optimization did not converge before'
                                                         'config.maxIter')
-        self.blendConvergenceFailedFlagKey = schema.addField('deblend_blendConvergenceFailedFlag', type='Flag',
-                                                         doc='at least one source in the blend'
-                                                             'failed to converge')
+        self.blendConvergenceFailedFlagKey = schema.addField('deblend_blendConvergenceFailedFlag',
+                                                             type='Flag',
+                                                             doc='at least one source in the blend'
+                                                                 'failed to converge')
         self.sourceConvergenceBitFlagKey = schema.addField('deblend_sourceConvergenceBitFlag', type=np.int32,
-                                                        doc="Flag for parameters that did not converge"
-                                                            "If this is zero, then all of the parameters"
-                                                            "of the source converged, otherwise it"
-                                                            "contains the bit flag for parameters that"
-                                                            "failed, which might differ depending on the"
-                                                            "source type")
+                                                           doc="Flag for parameters that did not converge"
+                                                               "If this is zero, then all of the parameters"
+                                                               "of the source converged, otherwise it"
+                                                               "contains the bit flag for parameters that"
+                                                               "failed, which might differ depending on the"
+                                                               "source type")
         self.edgePixelsKey = schema.addField('deblend_edgePixels', type='Flag',
                                              doc='Source had flux on the edge of the parent footprint')
         self.deblendFailedKey = schema.addField('deblend_failed', type='Flag',
--- a/python/lsst/meas/extensions/scarlet/deblend.py
+++ b/python/lsst/meas/extensions/scarlet/deblend.py
@@ -200,7 +200,7 @@ class ScarletDeblendConfig(pexConfig.Config):
 
     # Constraints
     sparse = pexConfig.Field(dtype=bool, default=True, doc="Make models compact and sparse")
-    morphThresh = pexConfig.Field(dtype=float, default=5,
+    morphThresh = pexConfig.Field(dtype=float, default=1,
                                   doc="Fraction of background RMS a pixel must have"
                                       "to be included in the initial morphology")
     monotonic = pexConfig.Field(dtype=bool, default=True, doc="Make models monotonic")
--- a/python/lsst/meas/extensions/scarlet/deblend.py
+++ b/python/lsst/meas/extensions/scarlet/deblend.py
@@ -113,12 +113,12 @@ def deblend(mExposure, footprint, config):
 
     Parameters
     ----------
-    mExposure: `MultibandExposure`
+    mExposure : `lsst.image.MultibandExposure`
         - The multiband exposure containing the image,
           mask, and variance data
-    footprint: `Footprint`
+    footprint : `lsst.detection.Footprint`
         - The footprint of the parent to deblend
-    config: `ScarletDeblendConfig`
+    config : `ScarletDeblendConfig`
         - Configuration of the deblending task
     """
     # Extract coordinates from each MultiColorPeak
@@ -128,7 +128,10 @@ def deblend(mExposure, footprint, config):
     images = mExposure.image[:, bbox].array
 
     # Use the inverse variance as the weights
-    weights = 1/mExposure.variance[:, bbox].array
+    if config.useWeights:
+        weights = 1/mExposure.variance[:, bbox].array
+    else:
+        weights = np.ones_like(images)
 
     # Use the mask plane to mask bad pixels and
     # the footprint to mask out pixels outside the footprint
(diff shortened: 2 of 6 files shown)


python/lsst/meas/extensions/scarlet/deblend.py
```diff
@@ -135,26 +141,29 @@ def deblend(mExposure, footprint, log, config):
...
-    else:
-        Source = LsstSource
-    sources = [
-        Source(frame=frame, peak=center, observation=observation, bgRms=bgRms,
-               bbox=bbox, symmetric=config.symmetric, monotonic=config.monotonic,
-               centerStep=config.recenterPeriod)
-        for center in footprint.peaks
-    ]
+    psfShape = (config.modelPsfSize, config.modelPsfSize)
+    model_psf = PSF(partial(gaussian, sigma=.8), shape=(None,)+psfShape)
```
taranu: Isn't that sigma=.8 important enough to set in the config with that default?

python/lsst/meas/extensions/scarlet/deblend.py
```diff
@@ -184,20 +197,25 @@ class ScarletDeblendConfig(pexConfig.Config):
...
 
     # Other scarlet paremeters
-    useWeights = pexConfig.Field(dtype=bool, default=False, doc="Use inverse variance as deblender weights")
     usePsfConvolution = pexConfig.Field(
         dtype=bool, default=True,
         doc=("Whether or not to convolve the morphology with the"
              "PSF in each band or use the same morphology in all bands"))
+    modelPsfSize = pexConfig.Field(
+        dtype=int, default=11,
+        doc="The model PSF will be (modelPsfSize, modelPsfSize)")
```
taranu: "Model PSF side length in pixels" or something like that?

python/lsst/meas/extensions/scarlet/deblend.py
```diff
@@ -330,6 +348,13 @@ def _addSchemaKeys(self, schema):
         self.blendConvergenceFailedKey = schema.addField('deblend_blendConvergenceFailed', type='Flag',
                                                          doc='at least one source in the blend'
                                                              'failed to converge')
+        self.sourceConvergenceFlagKey = schema.addField('deblend_sourceConvergenceFlag', type=np.int32,
+                                                        doc="Flag for parameters that did not converge"
+                                                            "If this is zero, then all of the parameters"
+                                                            "of the source converged, otherwise it"
+                                                            "contains the bit flag for parameters that"
+                                                            "failed, which might differ depending on the"
+                                                            "source type")
```
taranu: I find it somewhat confusing that deblend_blendConvergenceFailed is a flag whose name doesn't end in _flag (why not deblend_blendConvergence_flag?), whereas deblend_sourceConvergenceFlag isn't actually a flag (how about something like deblend_sourceConvergence_bits?).

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -20,179 +20,142 @@
...
+    """Initialize a Source
+
+    The user can specify the number of desired components
+    for the modeled source. If scarlet cannot initialize a
+    model with the desired number of components it continues
+    to attempt initialization of one fewer component until
+    it finds a model that can be initialized. If all of the
+    models fail, including a `PointSource` model, then skip
+    the source.
+
```
taranu, line 57: I realize one can and should refer to scarlet's source here but it would be helpful to give a brief hint/summary as to how/why source initialization might fail.

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -20,179 +20,142 @@
...
+    ymin = bbox.getMinY()
+    center = np.array([peak.getIy()-ymin, peak.getIx()-xmin], dtype=int)
+
+    while components > 1:
+        try:
+            source = MultiComponentSource(frame, center, observation, symmetric=symmetric,
+                                          monotonic=monotonic, thresh=thresh)
+            if (np.any([np.isnan(c.sed) for c in components]) or
+                    np.all([c.sed <= 0 for c in source.components])):
+                logger.warning("Could not initialize")
```
taranu, line 91: This warning isn't very helpful. Why isn't this caught upstream?

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -20,179 +20,142 @@
...
+                raise ValueError("Could not initialize source")
+        except Exception:
+            # If the MultiComponentSource failed to initialize
+            # try an ExtendedSource
+            components -= 1
```
taranu, line 96: Are none of the Exceptions raised by scarlet itself worth warning for?

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -20,179 +20,142 @@
...
+    else:
+        for p, parameter in enumerate(source.parameters):
+            if not parameter.converged:
+                converged += 2 << p
+    return converged
+
+
+def morphToHeavy(source, peakSchema, xy0=Point2I()):
+    """Convert the morphology to a `HeavyFootprint`
+    """
```
taranu, line 163: Where are the Parameter docs? (2)

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -20,179 +20,142 @@
...
-class LsstSource(ExtendedSource):
-    """LSST Base Source
 
-    It is most likely that a source in LSST might be slightly different
-    than the default scarlet sources, so this class allows us to define the
-    default initialization and update constraints for general sources in
-    LSST images.
+def checkConvergence(source):
+    """Check that a source converged
     """
```
taranu, line 134: Where are the Parameter docs? (1)

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -20,179 +20,142 @@
...
+    peakFlux = morph[cy, cx]
+    tfoot.addPeak(cx+xmin, cy+ymin, peakFlux)
+    timg = afwImage.ImageF(morph, xy0=xy0)
+    timg = timg[tfoot.getBBox()]
+    heavy = afwDet.makeHeavyFootprint(tfoot, afwImage.MaskedImageF(timg))
+    return heavy
+
+
+def modelToHeavy(source, filters, xy0=Point2I(), observation=None, dtype=np.float32):
+    """Convert the model to a `MultibandFootprint`
```
taranu, line 184: Where are the Parameter docs? (3)

python/lsst/meas/extensions/scarlet/deblend.py
```diff
@@ -112,21 +108,31 @@ def _computePsfImage(self, position=None):
     return psfImage
 
 
-def deblend(mExposure, footprint, log, config):
+def deblend(mExposure, footprint, config):
+    """Deblend a parent footprint
+
+    Parameters
+    ----------
+    mExposure: `MultibandExposure`
```
taranu: See https://developer.lsst.io/python/numpydoc.html#py-docstring-parameters
- Must have a space before and after the colon
- Parameter types should have the full namespace (`lsst.foo.bar`); I haven't checked to see if they work without them
- It would be nice to maintain consistency (always end with a period; only add hyphens for lists)

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -20,179 +20,142 @@
...
+
+    Parameters
+    ----------
+    frame: `LsstFrame`
+        The model frame for the scene
+    peak: `PeakRecord`
+        Record for a peak in the parent `PeakCatalog`
+    observation: `LsstObservation`
+        The images, psfs, etc, of the observed data.
+    bbox: `Rect`
```
taranu: I don't think you imported Rect anywhere so this probably won't work.

taranu (changes requested):
Mostly fine but I left a few comments that should be addressed.

python/lsst/meas/extensions/scarlet/source.py
fred3m: It should, because I don't think this actually creates a link to the class. Otherwise I'd have to import `Rect` but never use it, which would cause `flake8` to fail.

python/lsst/meas/extensions/scarlet/deblend.py
fred3m: `deblend_blendConvergenceFailed` -> `deblend_blendConvergenceFailedFlag`
`deblend_sourceConvergenceFlag` -> `deblend_sourceConvergenceBitFlag`

python/lsst/meas/extensions/scarlet/source.py
fred3m: clarified the docs

python/lsst/meas/extensions/scarlet/source.py
fred3m: This was part of a discussion that went on for weeks. scarlet used to throw a `SourceInitError` to inform the user that initialization failed but Peter felt strongly about having users check for a bad initialization and I finally got tired of fighting him on it.

python/lsst/meas/extensions/scarlet/source.py
fred3m: Not really. See previous comment, with the addendum that this happens quite often for faint sources and there is no need to overload the logger with useless warnings, since 90% of the time the warning would be "this source has a SNR too low to initialize as an `ExtendedSource`.

python/lsst/meas/extensions/scarlet/__init__.py
```diff
@@ -24,4 +24,4 @@
 from .observation import *
 from .source import *
 from .blend import *
-from .deblend import *
+from . import deblend
```
taranu, line 27: Missing from .version import * (the docs won't build without it)

python/lsst/meas/extensions/scarlet/source.py
taranu: I thought the whole point was to create links, but apparently it doesn't? I'll have to ask about that.

Anyway, in this case since you're not actually importing Rect, the full namespace would be more useful.

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -20,179 +20,197 @@
...
+    the joint coadd (all bands combined together into
+    single signal-to-noise weighted image for initialization)
+    to initialize, and a true spurious detection will not have
+    enough signal to initialize as a point source.
+    If all of the models fail, including a `PointSource` model,
+    then this source is skipped.
+
+    Parameters
+    ----------
+    frame: `LsstFrame`
```
taranu: Need a space before and after the colon here and in most of the added parameter docstrings.

Sadly I haven't been able to convince PyCharm to do this automatically.

python/lsst/meas/extensions/scarlet/source.py
fred3m: Ok. It's a good thing you had me do this, I forgot that the stack doesn't call it `Rect`, it's `Box2I`.

python/lsst/meas/extensions/scarlet/source.py
fred3m: This is an odd style guide requirement, as DM typically doesn't adhere to the spaces around operators rule. Numpydocs doesn't require this to compile properly, so this is just a waste of time (not your fault, I'm just venting). But I fixed it throughout the package, since I never use a space on both sides of the colon (because why would a sane person!).

fred3m:
Should be ready to go now.

PR #19, opened 2020-02-07
tickets/DM-23364

fred3m:
Allow sources on the edge with poorly constrained positions to shift. This ticket also marks sources on the edge to notify users downstream.

### DIFF
--- a/python/lsst/meas/extensions/scarlet/deblend.py
+++ b/python/lsst/meas/extensions/scarlet/deblend.py
@@ -37,7 +37,7 @@ import lsst.afw.image as afwImage
 import lsst.afw.detection as afwDet
 import lsst.afw.table as afwTable
 
-from .source import init_source, modelToHeavy
+from .source import initSource, modelToHeavy
 from .blend import LsstBlend, checkBlendConvergence
 from .observation import LsstFrame, LsstObservation
 
@@ -190,7 +190,7 @@ def deblend(mExposure, footprint, config):
         else:
             raise ValueError("Unrecognized sourceModel")
 
-        source = init_source(frame=frame, peak=center, observation=observation, bbox=bbox,
+        source = initSource(frame=frame, peak=center, observation=observation, bbox=bbox,
                              symmetric=config.symmetric, monotonic=config.monotonic,
                              thresh=config.morphThresh, components=components)
         if source is not None:
--- a/python/lsst/meas/extensions/scarlet/deblend.py
+++ b/python/lsst/meas/extensions/scarlet/deblend.py
@@ -22,6 +22,7 @@
 from functools import partial
 
 import numpy as np
+import scarlet
 from scarlet.psf import PSF, gaussian
 from scarlet import PointSource, ExtendedSource, MultiComponentSource
 
@@ -108,7 +109,7 @@ def _computePsfImage(self, position=None):
     return psfImage
 
 
-def getFootprintMask(footprint, mExposure, config):
+def getFootprintMask(footprint, mExposure):
     """Mask pixels outside the footprint
 
     Parameters
@@ -118,8 +119,6 @@ def getFootprintMask(footprint, mExposure, config):
           mask, and variance data
     footprint : `lsst.detection.Footprint`
         - The footprint of the parent to deblend
-    config : `ScarletDeblendConfig`
-        - Configuration of the deblending task
 
     Returns
     -------
@@ -159,7 +158,7 @@ def deblend(mExposure, footprint, config):
         weights = np.ones_like(images)
 
     # Mask out the pixels outside the footprint
-    mask = getFootprintMask(footprint, mExposure, config)
+    mask = getFootprintMask(footprint, mExposure)
     weights *= ~mask
 
     psfs = _computePsfImage(mExposure, footprint.getCentroid()).array.astype(np.float32)
@@ -191,8 +190,9 @@ def deblend(mExposure, footprint, config):
             raise ValueError("Unrecognized sourceModel")
 
         source = initSource(frame=frame, peak=center, observation=observation, bbox=bbox,
-                             symmetric=config.symmetric, monotonic=config.monotonic,
-                             thresh=config.morphThresh, components=components)
+                            symmetric=config.symmetric, monotonic=config.monotonic,
+                            thresh=config.morphThresh, components=components,
+                            edgeDistance=config.edgeDistance, shifting=False)
         if source is not None:
             sources.append(source)
         else:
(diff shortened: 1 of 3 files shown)


python/lsst/meas/extensions/scarlet/deblend.py
```diff
@@ -222,36 +222,22 @@ class ScarletDeblendConfig(pexConfig.Config):
                                          "iterations to exit fitter"))
 
     # Blend Configuration options
-    recenterPeriod = pexConfig.Field(dtype=int, default=5,
-                                     doc=("Number of iterations between recentering"))
-    exactLipschitz = pexConfig.Field(dtype=bool, default=True,
-                                     doc=("Calculate exact Lipschitz constant in every step"
-                                          "(True) or only calculate the approximate"
-                                          "Lipschitz constant with significant changes in A,S"
-                                          "(False)"))
```
natelust, line 231: These seem like they should be on their own commit, I dont see how it relates to the edge work

python/lsst/meas/extensions/scarlet/deblend.py
```diff
@@ -264,8 +250,6 @@ class ScarletDeblendConfig(pexConfig.Config):
     processSingles = pexConfig.Field(
         dtype=bool, default=False,
         doc="Whether or not to process isolated sources in the deblender")
-    storeHistory = pexConfig.Field(dtype=bool, default=False,
-                                   doc="Whether or not to store the history for each source")
```
natelust, line 268: same with all these changes

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -33,9 +33,46 @@
...
+        ):
+            return True
+    return False
+
+
 def initSource(frame, peak, observation, bbox,
-                symmetric=False, monotonic=True,
-                thresh=5, components=1):
+               symmetric=False, monotonic=True,
+               thresh=5, components=1, edgeDistance=1, shifting=False):
```
natelust, line 80: This is not wrong per-say, but keep in mind that as the number of function arguments grows, it becomes hard to keep track of and or call. At some length it is worth considering a dedicated class to set and pass.

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -116,6 +155,21 @@ def initSource(frame, peak, observation, bbox,
             # so skip this source
             return None
 
+    if hasEdgeFlux(source, edgeDistance):
+        # The detection algorithm implemented in meas_algorithms
+        # does not place sources within the edge mask
+        # (roughly 5 pixels from the edge). This results in poor
+        # deblending of the edge source, which for bright sources
+        # may ruin an entire blend. So we reinitialize edge sources
+        # to allow for shifting and return the result.
```
natelust: Note that while shifting may be defined w/n scarlet, it does not seem to be defined in this package, which makes this comment less useful. Consider either a brief description here (and or somewhere else in this package) or a link to scarlet docs where shifting is talked about.

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -33,9 +33,46 @@
...
+        Whether or not the source has flux on the edge.
+    """
+    assert edgeDistance > 0
+
+    # Use the first band that has a non-zero SED
+    band = np.min(np.where(source.sed > 0)[0])
+    model = source.get_model()[band]
+    for edge in range(edgeDistance):
+        if (
+            np.any(model[edge-1] > 0) or
```
natelust, line 69: are scarlet models somehow 1 based indexing? if edge in the loop is zero, 0-1 is -1 and so this will test the last pixel, then the 0th etc. Is this what you were intending?

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -33,9 +33,46 @@
 logger = lsst.log.Log.getLogger("meas.deblender.deblend")
 
 
+def hasEdgeFlux(source, edgeDistance=1):
+    """hasEdgeFlux
+
+    Determine whether or not a source has flux within `edgeDistance`
+    of the edge.
```
natelust, line 40: Are these the edges of a box containing the entire footprint to deblend, or a box only containing an individual source, i.e. a smaller view?

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -33,9 +33,46 @@
...
+    Returns
+    -------
+    isEdge: `bool`
+        Whether or not the source has flux on the edge.
+    """
+    assert edgeDistance > 0
+
+    # Use the first band that has a non-zero SED
+    band = np.min(np.where(source.sed > 0)[0])
+    model = source.get_model()[band]
```
natelust, line 66: if you did something like:
```
edges = np.ones(model.shape)
edges[edgeDistance+1:-edgeDistance-1, edgeDistance+1:-edgeDistance-1] = 0
return np.sum(model*edges)>0
```
you could avoid all the ifs, and reduce the number of implicit loops

natelust (approved):

python/lsst/meas/extensions/scarlet/deblend.py
fred3m: Agreed. I came across obsolete parameters when I was adding the edge ones and rather than forget to remove them I deleted them right away. But it seems like a small enough change to not warrant the time that it would take to pick lines for a separate commit.

python/lsst/meas/extensions/scarlet/deblend.py
fred3m: same explanation

python/lsst/meas/extensions/scarlet/source.py
fred3m: Yeah, I think this is just a stylistic question. I only prefer using a class if the majority of the components are going to be used in multiple function calls. In this case, while there are a lot of arguments I'm not a big fan of using a class that is only called once in a codebase.

python/lsst/meas/extensions/scarlet/source.py
fred3m: Note [the assert](https://github.com/lsst/meas_extensions_scarlet/blob/tickets/DM-23364/python/lsst/meas/extensions/scarlet/source.py#L57). So the minimum value of `edgeDistance` is 1, which describes the edge.

I updated the docstring to make this clear.

python/lsst/meas/extensions/scarlet/source.py
fred3m: Good question. `source.get_model` returns an array that is the size of the blend, with the source injected into it. So this only tests against a true image edge.

python/lsst/meas/extensions/scarlet/source.py
fred3m: True, but for N = length of a side, that code scales as N^2 whereas the current algorithm scales with N. I verified this in a jupyter notebook.

PR #24, opened 2020-05-25
tickets/DM-24862

fred3m:
Update to latest proxmin and scarlet.


python/lsst/meas/extensions/scarlet/deblend.py
```diff
@@ -39,14 +39,20 @@
...
-from .blend import LsstBlend, checkBlendConvergence
-from .observation import LsstFrame, LsstObservation
 
 __all__ = ["deblend", "ScarletDeblendConfig", "ScarletDeblendTask"]
 
 logger = lsst.log.Log.getLogger("meas.deblender.deblend")
 
 
+def _checkBlendConvergence(blend, f_rel):
+    """Check whether or not a blend has converged
```
taranu, line 49: Needs numpy docs for args and returns (even if they're fairly straightforward)

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -101,12 +143,11 @@ def initSource(frame, peak, observation, bbox,
     ----------
     frame : `LsstFrame`
         The model frame for the scene
-    peak : `PeakRecord`
-        Record for a peak in the parent `PeakCatalog`
-    observation : `LsstObservation`
-        The images, psfs, etc, of the observed data.
-    bbox : `lsst.geom.Box2I`
-        The bounding box of the parent footprint.
+    center : `tuple`
```
taranu: You could put `tuple` [`float`]. I'm not sure if numpydoc allows specifying the length exactly though.

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -115,45 +156,85 @@ def initSource(frame, peak, observation, bbox,
...
         each pixel in the initialization
-    components : int
-        The number of components for the source.
-        If `components=0` then a `PointSource` model is used.
+    maxComponents : int
+        The maximum number of components in a source.
+        If a source fails to initialize with `maxComponents` then it
+        will continue to subtract one from the number of components
+        until it reaches zero (which fits a point source).
+        If a point source cannot be fit then the source is skipped.
```
taranu, line 165: The docstring for fallback implies that this will only happen if fallback = True, so you should probably mention that here.

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -115,45 +156,85 @@ def initSource(frame, peak, observation, bbox,
...
+        If `edgeDistance` is `None` then the edge check is ignored.
+    shifting : bool
+        Whether or not to fit the position of a source.
+        This is an expensive operation and is typically only used when
+        a source is on the edge of the detector.
+    downgrade : bool
+        Whether or not to decrease the number of components for sources
+        with small bounding boxes. For example, a source with no flux
+        outside of its 16x16 box is unlikely to be resolved enough
+        for multiple components, so a single source can be used.
```
taranu, line 180: Maybe just explicitly state how downgrade lowers maxcomponents? You can explain the logic in a Notes section.

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -115,45 +156,85 @@ def initSource(frame, peak, observation, bbox,
...
         try:
             source = MultiComponentSource(frame, center, observation, symmetric=symmetric,
                                           monotonic=monotonic, thresh=thresh, shifting=shifting)
             if (np.any([np.any(np.isnan(c.sed)) for c in source.components])
                     or np.any([np.all(c.sed <= 0) for c in source.components])
                     or np.any([np.any(~np.isfinite(c.morph)) for c in source.components])):
-                logger.warning("Could not initialize")
-                raise ValueError("Could not initialize source")
-            if hasEdgeFlux(source, edgeDistance):
+                msg = "Could not initlialize source at {} with {} components".format(center, maxComponents)
```
taranu: initialize

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -247,7 +323,29 @@ def modelToHeavy(source, filters, xy0=Point2I(), observation=None, dtype=np.floa
...
     """
     if observation is not None:
-        model = observation.render(source.get_model()).astype(dtype)
+        # We want to convolve the model with the observed PSF,
+        # which means we need to grow the model box by the PSF to
+        # account for all of the flux after convolution.
+        # FYI: The `scarlet.Box` class implements the `&` operator
+        # to take the intersection of two boxes.
+
+        # Get the PSF size and radii to grow the
```
taranu: ... to grow the box by?

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -247,7 +323,29 @@ def modelToHeavy(source, filters, xy0=Point2I(), observation=None, dtype=np.floa
...
+        bbox = Box(shape, origin=origin)
+        # Only use the portion of the convolved model that fits in the image
+        overlap = bbox & source.model_frame
+        # Load the full multiband model in the larger box
+        model = source.model_to_frame(overlap)
+        # Convolve the model with the PSF in each band
+        # Always use a real space convolution to limit artifacts
+        model = observation.convolve(model, convolution_type="real").astype(dtype)
+        # Update xy0 with the origin of the sources box
+        xy0 = Point2I(overlap.origin[-1]+xy0.x, overlap.origin[-2]+xy0.y)
```
taranu: The style guide says + and - should have spaces, so I'd suggest staying consistent with that (there are spaces just above).

taranu (commented):
Looks fine; I just have minor suggestions/fixes.

python/lsst/meas/extensions/scarlet/deblend.py
fred3m: I don't think this is true for private methods. It's certainly common practice to not include arguments and parameters for private methods and the only place this is mentioned in the dev guid is

```
def _private(self):
        """Do something internally.

        By default private members are not included in the HTML docs either.

        Notes
        -----
        Private members are any methods or attributes that start with an
        underscore and are *not* special. By default they are not included in
        the output.

        However, you should still provide docstrings for private members to
        document code for internal developers.
        """
        pass
```

Looking at the other method in that section (https://developer.lsst.io/python/numpydoc.html#complete-example-module) I feel like this implies that docstrings are needed but in a case where the parameters are obvious we don't need to be so verbose. But if you can find somewhere that says otherwise...

python/lsst/meas/extensions/scarlet/source.py
fred3m: I don't want to be too explicit because the exact implementation might change over time and I don't want the docs and the code to get out of sync.

PR #26, opened 2020-09-23
tickets/DM-26647

fred3m:
This coadd adds a number of new columns that are useful in tracking
down errors or bugs in the deblender. Included in these is the
"deblend_error" column, which keeps track of errors caused by
a coadd that is smaller than the coadds in the other bands and
cannot load a PSF.

### DIFF
--- a/python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
+++ b/python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
@@ -122,7 +122,7 @@ def _computePsfImage(self, position=None):
     if not isinstance(position, Point2D) and position is not None:
         position = Point2D(position[0], position[1])
 
-    for band, single in enumerate(self.singles):
+    for bidx, single in enumerate(self.singles):
         try:
             if position is None:
                 psf = single.getPsf().computeImage()
@@ -137,7 +137,7 @@ def _computePsfImage(self, position=None):
             # data in some bands, so we track this error to distinguish it
             # from unknown errors.
             msg = "Failed to compute PSF at {} in band {}"
-            raise IncompleteDataError(msg.format(position, self.filters[band]))
+            raise IncompleteDataError(msg.format(position, self.filters[bidx]))
 
     left = np.min([psf.getBBox().getMinX() for psf in psfs])
     bottom = np.min([psf.getBBox().getMinY() for psf in psfs])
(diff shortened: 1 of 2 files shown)


python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -98,13 +122,23 @@ def _computePsfImage(self, position=None):
...
         position = Point2D(position[0], position[1])
 
-    for single in self.singles:
-        if position is None:
-            psf = single.getPsf().computeImage()
-            psfs.append(psf)
-        else:
-            psf = single.getPsf().computeImage(position)
-            psfs.append(psf)
+    for band, single in enumerate(self.singles):
```
taranu: I recommend consistently naming iteration indices idx_ or something that indicates that it's an index/integer since band is often a string.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -98,13 +122,23 @@ def _computePsfImage(self, position=None):
...
+        try:
+            if position is None:
+                psf = single.getPsf().computeImage()
+                psfs.append(psf)
+            else:
+                psf = single.getPsf().computeImage(position)
+                psfs.append(psf)
+        except InvalidParameterError:
```
taranu, line 133: I'm a bit confused as to why the try block doesn't throw an IncompleteDataError itself if it's always due to a missing PSF model?

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -564,24 +619,56 @@ def deblend(self, mExposure, sources):
                 # that the peaks stay consistent
                 for k in skipped:
                     sources.insert(k, None)
+            # Catch the errors that we know about
+            except IncompleteDataError:
+                blendError = "IncompleteDataError"
```
taranu: This seems unnecessary; you could replace it with blendError = `type(e).__name__` in the except part below and test for blendError in {"IncompleteDataError", "ScarletGradientError"} if you want to check against a set of possible/expected errors.

taranu (commented):
Just a few minor questions/suggestions.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: The `InvalidParameterError` is thrown by the `Exposure.getPsf` method, which is defined in `afw`. I could just allow that to fail and catch it in `ScarletDeblendTask.deblend` but in general I like to isolate the code that causes a known issue in case an issue somewhere else arises later with the same error type.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: Ok, I cleaned up the code a bit as per this suggestion.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -619,24 +619,22 @@ def deblend(self, mExposure, sources):
...
-            # Catch all unknown errors to prevent the taks from crashing,
-            # but store the result
+            # Catch all errors and filter out the ones that we know about
             except Exception as e:
-                if self.config.catchFailures:
+                blendError = type(e).__name__
+                if isinstance(e, ScarletGradientError):
+                    src.set(self.iterKey, e.iterations)
+                elif not isinstance(e, IncompleteDataError):
                     blendError = "UnknownError"
```
taranu, line 628: Why override blendError instead of setting it to e.g. `f'{Unexpected {blendError}'`?

taranu (approved):
Looks fine; just one optional suggestion.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: 2 reasons:
1. I don't know the length of the name of the unexpected error that is raised, and unlike numpy the stack isn't smart enough to just chop off a string that is too long, so if the raised error is longer than the number of characters I've allotted for that column in the table it will raise an Exception. I could use `error = ("UnknownError:" + blendError)[:25]` except for the next issue...

2. I want to be able to search for `catalog["deblend_error"] == "UnknownError"` to filter for sources with unknown errors. It is true that `(catalog["deblend_error"] != "ScarletGradientError") & (catalog["deblend_error"] != "IncompleteDataError")` would also work, but as time goes on I expect that we will identify more known failure modes and I'd prefer to not have to change all of my scripts each time. No matter what I'll want to investigate all of the new "UnknownError" blends, so I'm good with this the way it is.

PR #28, opened 2020-10-28
tickets/DM-27208

fred3m:
Add columns to make it easier to cross check meas_deblender and meas_extensions_scarlet.

### DIFF
--- a/python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
+++ b/python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
@@ -762,9 +762,9 @@ class ScarletDeblendTask(pipeBase.Task):
         source.set(self.runtimeKey, 0)
         # Set the center of the parent
         bbox = fp.getBBox()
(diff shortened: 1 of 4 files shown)


taranu (approved):

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -592,10 +596,16 @@ def deblend(self, mExposure, sources):
                 self._skipParent(src, mask)
                 self.log.trace('Parent %i: skipping masked footprint', int(src.getId()))
                 continue
-            if len(peaks) > self.config.maxNumberOfPeaks:
+            if self.config.maxNumberOfPeaks > 0 and len(peaks) > self.config.maxNumberOfPeaks:
```
taranu, line 599: It doesn't particularly matter much but you could store this value before the loop rather than checking every time, or simply set maxNumberOfPeaks to np.inf if it's zero.

PR #33, opened 2021-02-05
DM-28542: Implement RFC-750

fred3m:
The deblend_parentNPeaks column was never updated, and there was no deblend_parentNChild column to keep track of the number of children actually deblended from a parent. So this PR fixes those erroneous columns.

### DIFF
--- a/python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
+++ b/python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
@@ -525,6 +525,9 @@ class ScarletDeblendTask(pipeBase.Task):
         self.parentNPeaksKey = schema.addField("deblend_parentNPeaks", type=np.int32,
                                                doc="Same as deblend_n_peaks, but the number of peaks "
                                                    "in the parent footprint")
+        self.parentNChildKey = schema.addField("deblend_parentNChild", type=np.int32,
+                                               doc="Same as deblend_nChild, but the number of "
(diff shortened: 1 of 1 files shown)


python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -525,6 +537,9 @@ def _addSchemaKeys(self, schema):
         self.parentNPeaksKey = schema.addField("deblend_parentNPeaks", type=np.int32,
                                                doc="Same as deblend_n_peaks, but the number of peaks "
                                                    "in the parent footprint")
+        self.parentNChildKey = schema.addField("deblend_parentNChild", type=np.int32,
+                                               doc="Same as deblend_nChild, but the number of "
+                                                   "deblended children from the parent footprint")
```
laurenam: I get the point of drawing an analogy to `deblend_nChild`, but perhaps it would be better to leave that as an afterthought as opposed to the lead-in (especially given how far up one has to look to find the doc for `deblend_nChild`)?  To fill in the blank, it does read rather strange to me:
"Same as "Number of children this object has (defaults to 0)", but the number of deblended children from the parent footprint"

laurenam:
In principle, this PR should be titled (https://developer.lsst.io/work/flow.html#make-a-pull-request):
DM-28542: Implement RFC-750

laurenam (approved):

PR #37, opened 2021-02-24
DM-28947: Add option to optionally set spectra init

fred3m:


python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -240,6 +240,15 @@ def deblend(mExposure, footprint, config):
...
     centers = [np.array([peak.getIy()-ymin, peak.getIx()-xmin], dtype=int) for peak in footprint.peaks]
 
+    # Choose whether or not to use the improved spectral initialization
+    if config.setSpectra:
+        if config.maxSpectrumCutoff == -1:
+            spectrumInit = True
+        else:
+            spectrumInit = len(centers) * bbox.getArea() < config.maxSpectrumCutoff
+    else:
+        spectrumInit = config.setSpectra
```
erykoff: I think in this case this is always `False`, right?  I think it should just be set to `False`.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -240,6 +240,15 @@ def deblend(mExposure, footprint, config):
     ymin = bbox.getMinY()
     centers = [np.array([peak.getIy()-ymin, peak.getIx()-xmin], dtype=int) for peak in footprint.peaks]
 
+    # Choose whether or not to use the improved spectral initialization
+    if config.setSpectra:
+        if config.maxSpectrumCutoff == -1:
```
erykoff: Would it be simpler here (and below) to say that anything `<= 0` turns this off?

erykoff (approved):
A couple of minor comments that I think will make it a bit easier to read/use.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: Yes. Originally I tried to use `True` to always use `setSpectra`, `False` to never use it, and `None` to check `maxSpectrumCutoff`, but the stack won't let me use `None` as the default for a `bool` (which makes sense, but :( ).

erykoff:
One more thing: the title of the PR has a typo.

PR #40, opened 2021-04-08
DM-29633: Fix log level issues with scarlet and proxmin

fred3m:


python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -40,19 +40,14 @@
...
-scarletInitLogger.setLevel(logging.ERROR)
-scarletSourceLogger.setLevel(logging.ERROR)
+# Scarlet and proxmin have a different definition of log levels than the stack,
+# so even "warnings" occur far more often than we would like.
+# So for now we only display scarlet and proxmin errors, as all other
+# scarlet outputs would be considered "TRACE" by our standards.
+scarletLogger = logging.getLogger("scarlet")
+scarletLogger.setLevel(logging.ERROR)
+proxminLogger = logging.getLogger("proxmin")
+proxminLogger.setLevel(logging.ERROR)
```
erykoff, line 50: Just to be clear, the change here is (a) to apply these log levels to all scarlet logs not just initialisation [sic] and source; (b) apply these changes to proxmin which wasn't done before?

erykoff (approved):
Looks good, looking for a clarification on the change though.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: Correct. I originally thought that all of the offending scarlet logs were just contained in the initialization module, but it turns out that other modules have useless logs (for a large survey) too.

And (c) remove a print statement in meas_extensions_scarlet that was probably used for debugging at some point.

erykoff: (And now I see you wrote this on the jira ticket).  Thanks!

PR #43, opened 2021-08-31
DM-31528: Add more log messages

arunkannawadi:
Scarlet can be silent for too long when deblending sources, making it look like the job is hung. This PR allows a user to request more frequent logging (at VERBOSE level) to monitor the progress.

### DIFF
--- a/python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
+++ b/python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
@@ -454,6 +454,11 @@ class ScarletDeblendConfig(pexConfig.Config):
         doc="Names of flags which should never be deblended."
     )
 
+    # Logging option(s)
+    loggingInterval = pexConfig.Field(
+        dtype=int, default=600,
+        doc="Interval (in seconds) to log messages (at VERBOSE level) while deblending sources."
+    )
     # Testing options
     # Some obs packages and ci packages run the full pipeline on a small
     # subset of data to test that the pipeline is functioning properly.
@@ -675,6 +680,7 @@ class ScarletDeblendTask(pipeBase.Task):
 
         filters = mExposure.filters
         self.log.info("Deblending %d sources in %d exposure bands", len(catalog), len(mExposure))
+        nextLogTime = time.time() + self.config.loggingInterval
 
         # Add the NOT_DEBLENDED mask to the mask plane in each band
         if self.config.notDeblendedMask:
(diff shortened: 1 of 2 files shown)


python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -674,7 +679,8 @@ def deblend(self, mExposure, catalog):
             idFactory.notify(maxId)
 
         filters = mExposure.filters
-        self.log.info(f"Deblending {len(catalog)} sources in {len(mExposure)} exposure bands")
+        self.log.info("Deblending %d sources in %d exposure bands", len(catalog), len(mExposure))
+        logTime = time.time()
```
timj: maybe call this `lastLoggedTime`? (or see later comment about changing it to next time a log message is needed)

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -827,6 +833,10 @@ def deblend(self, mExposure, catalog):
                     catalog=catalog,
                     scarletSource=scarletSource,
                 )
+            # Log a message if it has been a while since the last log.
+            if (currentTime := time.time()) > logTime + self.config.loggingInterval:
```
timj: Rather than always comparing against `logTime + loggingInterval` every time, instead use `nextLogTime` instead of `lastLoggedTime` and calculate the next time at which you want to log and then simplify the if statement and only calculate the time with offset once you do log.

timj (approved):
Looks great. One minor comment about when the 600 seconds is added onto the time.

fred3m (approved):
This looks good to me. I never thought I would see the day when people thought that scarlet didn't have _enough_ logs. :)

I do wonder if this will completely fix the problem. Sometimes there can be _very_ large blends, for example there are some HSC patches where a single blend takes up almost the entire patch due to satellite trails (unless that has been fixed recently), so it is possible that those will still time out with this fix and we might need to put a similar heartbeat into scarlet itself.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
arunkannawadi: That's an excellent suggestion! Adopted this now.

arunkannawadi:
Thanks Fred! Yes, it would be good to have a similar heartbeat log message in scarlet itself. This fix implicitly assumes that any single parent can be deblended on timescales that are much smaller than `loggingInterval`. Perhaps, scarlet could take in this parameter as an input argument from this task itself?

PR #46, opened 2021-11-08
DM-31801: Use lsst prefix for lsst loggers

timj:


python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -52,7 +50,7 @@
 
 __all__ = ["deblend", "ScarletDeblendConfig", "ScarletDeblendTask"]
 
-logger = lsst.log.Log.getLogger("meas.deblender.deblend")
+logger = logging.getLogger("lsst.meas.deblender.deblend")
```
fred3m: Wow, looks like this has had the wrong label ever since we ported this module from `lsst.meas.deblender.deblend.py`. The correct logging name should be `lsst.meas.extensions.scarlet.scarletDeblendTask`, so you should just be able to change this to `logging.getLogger("__name__")` like you did for the `source` module.

fred3m (changes requested):
Please change the name of the logger for scarletDeblendTask, then this will be all set.

fred3m (approved):

PR #47, opened 2022-01-26
tickets/DM-32436: implement scarlet lite

fred3m:


python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -328,6 +331,136 @@ def deblend(mExposure, footprint, config):
...
+    # Extract coordinates from each MultiColorPeak
+    bbox = footprint.getBBox()
+
+    # Create the data array from the masked images
+    images = mExposure.image[:, bbox].array
+    variance = mExposure.variance[:, bbox].array
+
+    # Use the inverse variance as the weights
+    if config.useWeights:
+        weights = 1/mExposure.variance[:, bbox].array
```
taranu, line 356: Is the variance plane guaranteed to be non-nan? Don't you still want to set masked pixels to zero weight, or is the variance conveniently infinite for them already?

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -328,6 +331,136 @@ def deblend(mExposure, footprint, config):
...
+        weights = np.ones_like(images)
+        badPixels = mExposure.mask.getPlaneBitMask(config.badMask)
+        mask = mExposure.mask[:, bbox].array & badPixels
+        weights[mask > 0] = 0
+
+    # Mask out the pixels outside the footprint
+    mask = getFootprintMask(footprint, mExposure)
+    weights *= ~mask
+
+    psfs = _computePsfImage(mExposure, footprint.getCentroid()).array.astype(np.float32)
```
taranu, line 367: Why float32?

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -328,6 +331,136 @@ def deblend(mExposure, footprint, config):
...
+        parameterization = partial(
+            lite.init_adaprox_component,
+            bg_thresh=config.backgroundThresh,
+            max_prox_iter=config.maxProxIter,
+        )
+    elif config.optimizer == "fista":
+        parameterization = partial(
+            lite.init_fista_component,
+            bg_thresh=config.backgroundThresh,
+        )
```
taranu, line 420: No else clause?

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -328,6 +331,136 @@ def deblend(mExposure, footprint, config):
...
+    for k, center in enumerate(centers):
+        # This is just to make sure that there isn't a coding bug
+        if len(sources[k].components) > 0 and np.any(sources[k].center != center):
+            raise ValueError("Misaligned center, expected {center} but got {sources[k].center}")
+        # Store the record for the peak with the appropriate source
+        sources[k].detectedPeak = footprint.peaks[k]
+
+    blend = lite.LiteBlend(sources, observation)
+
+    # Initialize each source with its best fit spectrum
```
taranu, line 435: What is the 'best fit spectrum'?

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -328,6 +331,136 @@ def deblend(mExposure, footprint, config):
...
+    # This significantly cuts town on the number of iterations
+    # that the optimizer needs and usually results in a better
+    # fit, but using least squares on a very large blend causes memory issues.
```
taranu, line 438: Why does this result in a better fit? Also, I would have thought that the spectral degrees of freedom are basically irrelevant compared to the morphology and shouldn't impact runtime that much.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -328,6 +331,136 @@ def deblend(mExposure, footprint, config):
...
+
+    # Set the sources that could not be initialized and were skipped
+    skipped = [src for src in sources if src.is_null]
+
+    try:
+        blend.fit(max_iter=config.maxIter, e_rel=config.relativeError, min_iter=config.minIter)
+    except ArithmeticError:
+        # This occurs when a gradient update produces a NaN value
+        # This is usually due to a source initialized with a
+        # negative SED or no flux, often because the peak
```
taranu: Can you force sources to have small, positive fluxes? It seems like a shame to throw out a whole blend for one bad, faint source even if it's a false positive.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -339,16 +472,62 @@ class ScarletDeblendConfig(pexConfig.Config):
     - Other: Parameters that don't fit into the above categories
     """
     # Stopping Criteria
+    minIter = pexConfig.Field(dtype=int, default=1,
+                              doc="Minimum number of iterations before the optimizer is allwed to stop.")
```
taranu: allowed

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -339,16 +472,62 @@ class ScarletDeblendConfig(pexConfig.Config):
...
+                                         "flux re-distributed models and `1e-4` when making measurements "
+                                         "on the models themselves."))
 
     # Constraints
     morphThresh = pexConfig.Field(dtype=float, default=1,
                                   doc="Fraction of background RMS a pixel must have"
                                       "to be included in the initial morphology")
+    # Lite Parameters
+    # All of these parameters (except version) are only valid if version='lite'
+    version = pexConfig.Field(
```
taranu: version, optimizer and morphImage should all be ChoiceFields.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -771,17 +964,15 @@ def deblend(self, mExposure, catalog):
             try:
                 t0 = time.time()
```
taranu: The internet suggests time.monotonic or time.perf_counter, in case of system time changes (doubt that comes up much, but anyway).

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -807,7 +998,10 @@ def deblend(self, mExposure, catalog):
                 continue
 
             # Update the parent record with the deblending results
-            logL = blend.loss[-1]-blend.observations[0].log_norm
+            if self.config.version == "scarlet":
+                logL = -blend.loss[-1]+blend.observations[0].log_norm
```
taranu: There should be spaces around the plus: https://developer.lsst.io/python/style.html#binary-operators-should-be-surrounded-by-a-single-space-except-for
(that goes for other lines in this file too)

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -146,3 +146,78 @@ def modelToHeavy(source, mExposure, blend, xy0=Point2I(), dtype=np.float32):
     model = MultibandImage(mExposure.filters, model, valid.getBBox())
     mHeavy = MultibandFootprint.fromImages(mExposure.filters, model, footprint=foot)
     return mHeavy
+
+
+def liteModelToHeavy(source, mExposure, blend, xy0=Point2I(), dtype=np.float32, useFlux=False):
+    """Convert a scarlet model to a `MultibandFootprint`.
+    Parameters
+    ----------
+    source : `scarlet.Component`
```
taranu: Not a scarlet.Source?

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -146,3 +146,78 @@ def modelToHeavy(source, mExposure, blend, xy0=Point2I(), dtype=np.float32):
     model = MultibandImage(mExposure.filters, model, valid.getBBox())
     mHeavy = MultibandFootprint.fromImages(mExposure.filters, model, footprint=foot)
     return mHeavy
+
+
+def liteModelToHeavy(source, mExposure, blend, xy0=Point2I(), dtype=np.float32, useFlux=False):
```
taranu: Even if you don't mutate `xy0` currently, this should still be `xy0=None` and conditionally initialized to `Point2I()` below (https://developer.lsst.io/python/style.html#a-mutable-object-must-not-be-used-as-a-keyword-argument-default). Also, if you don't actually need a Point2I, you could just pass x0 and y0 floats.

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -146,3 +146,78 @@ def modelToHeavy(source, mExposure, blend, xy0=Point2I(), dtype=np.float32):
...
+    dh = py // 2
+    dw = px // 2
+
+    if useFlux:
+        source_box = source.flux_box
+        shape = source_box.shape
+        origin = source_box.origin
+    else:
+        source_box = source.bbox
+        shape = (source_box.shape[0], source_box.shape[1] + py, source_box.shape[2] + px)
```
taranu: Don't you only need to grow the box by `2*dh`, `2*dw` instead of py, px (if the psf shape is odd)?

taranu (changes requested):
A few issues and some requests for clarification.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: The variance is non-zero except in the mask, which is set to zero at line 365.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: For reference this entire function is copied from `deblend`, with a modified API for scarlet lite. But you're right, here it should be cast to whatever type `images` is, which in the science pipelines is float 32 but not guaranteed.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: 😞

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: It's a least squares fit that uses the morphology of each source to simultaneously fit their initial spectra. This has a huge impact on the number of iterations and overall final logL. But because it requires all of the sources to be projected on the full box of the blend, for large blends and/or dense regions there is a memory problem fitting them simultaneously.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: The reason that it impacts runtime is because the ADAM optimizer struggles if the initial amplitude is off by too much. So fitting the SEDs to have the correct amplitude does (unfortunately) make a big difference, at least for a model with as many degrees of freedom as ours.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: Thanks for pointing that out, this is an example of a comment that has bit rotted and why I don't like too many comments in general. In main scarlet we do threshold at 10^-20 and in scarlet lite we go one step further and any sources that could not be initialized to the data are initialized as a faint PSF at the center location. So we should be able to remove this entirely now.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: Thanks, I didn't know that was a thing.

python/lsst/meas/extensions/scarlet/source.py
fred3m: Good catch, I think that you're right.

python/lsst/meas/extensions/scarlet/source.py
fred3m: Isn't Point2I immutable?

fred3m: Oh, apparently not. Yes, I'll fix this.

fred3m:
Ok, I think that I addressed everything now. Once approved I'll run Jenkins before merging.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
taranu: My concern was more why aren't you using float64 for intermediate computation, even if you end up storing outputs as float32, unless you're really memory constrained.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
taranu: Ok, so you do a linear least squares fit for fluxes first, yes? If it's memory-constrained, can you use sparse matrices?

taranu (approved):

PR #49, opened 2022-04-04
DM-33690: Skip sources that are large with a lot of children

fred3m:


python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -592,13 +592,21 @@ class ScarletDeblendConfig(pexConfig.Config):
 
     # Size restrictions
     maxNumberOfPeaks = pexConfig.Field(
-        dtype=int, default=0,
+        dtype=int, default=200,
         doc=("Only deblend the brightest maxNumberOfPeaks peaks in the parent"
              " (<= 0: unlimited)"))
     maxFootprintArea = pexConfig.Field(
-        dtype=int, default=1000000,
+        dtype=int, default=int(1e5),
```
erykoff: For large integers like this, it's convenient to take advantage of the fact that `_`s are allowed in python ints.  So this can be `100_000` which I think is easier to read than either of the above options.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -592,13 +592,21 @@ class ScarletDeblendConfig(pexConfig.Config):
...
         doc=("Maximum area for footprints before they are ignored as large; "
              "non-positive means no threshold applied"))
+    maxAreaTimesPeaks = pexConfig.Field(
+        dtype=int, default=int(1e7),
```
erykoff: Similarly, this can be `10_000_000`.  Also, I think the default should be on a new line.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -1136,6 +1144,9 @@ def _isLargeFootprint(self, footprint):
             axes = afwEll.Axes(footprint.getShape())
             if axes.getB() < self.config.minFootprintAxisRatio*axes.getA():
                 return True
+        if self.config.maxAreaTimesPeaks > 0:
+            if footprint.getBBox().getArea() * len(footprint.peaks) > self.config.maxAreaTimesPeaks:
+                return True
```
erykoff, line 1147: Please update the docstring above with additional uses of `_isLargeFootprint`.

erykoff (approved):
Looks good, with some small comments.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: TIL that is allowed!

PR #50, opened 2022-04-11
DM-30108: Implement merging sources and flagging saturated

fred3m:


python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -452,6 +452,97 @@ def deblend_lite(mExposure, footprint, config, wavelets=None):
     return blend, skipped, spectrumInit
 
 
+class DeblenderMetrics:
+    """Metrics and measurements made on single sources.
+
+    This class exists so that all metrics calculated in the deblender can
+    be added as a single attribute to a scarlet source before it is
+    converted into a `SourceRecord`.
+    When DM-34414 this class will be eliminated and the metrics will
```
taranu: "[When DM-34414] is finished..."

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -452,6 +452,97 @@ def deblend_lite(mExposure, footprint, config, wavelets=None):
     return blend, skipped, spectrumInit
 
 
+class DeblenderMetrics:
+    """Metrics and measurements made on single sources.
+
+    This class exists so that all metrics calculated in the deblender can
```
taranu: This is nitpicky but you should omit redundant phrases like "This class exists so that..." and just begin like "Store deblender metrics to be added as attributes to a Scarlet source..." (channeling the [numpydoc section of the dev guide](https://developer.lsst.io/python/numpydoc.html#extended-summary)).

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -452,6 +452,97 @@ def deblend_lite(mExposure, footprint, config, wavelets=None):
...
+    converted into a `SourceRecord`.
+    When DM-34414 this class will be eliminated and the metrics will
+    be added to the schema using a pipeline task that calculates them
+    from the stored deconvolved models.
+    """
+    def __init__(self, maxOverlap, fluxOverlap, overlapRatio, blendedness):
+        """Initialize the class
+
+        All of the parameters are one dimensional numpy arrays,
+        with an element for each band in the observed images.
```
taranu: Seems like it should be a dict by band of a dataclass/pex Config with those properties, then.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -452,6 +452,97 @@ def deblend_lite(mExposure, footprint, config, wavelets=None):
...
+
+        Parameters
+        ----------
+        maxOverlap : `numpy.ndarray`
+            The maximum overlap that the source has with its neighbors.
+            This is useful as a metric for determining how blended a source
```
taranu: Explanatory text should go in the notes section below the one-sentence param doc.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -452,6 +452,97 @@ def deblend_lite(mExposure, footprint, config, wavelets=None):
...
+            is because if it only overlaps with other sources at or below
+            the noise level, it is likely to be a mostly isolated source
+            in the deconvolved model frame.
+        fluxOverlap : `numpy.ndarray`
+            The total flux from neighbors overlapping with the current source.
+        overlapRatio : `numpy.ndarray`
+            This is the fraction of `flux from neighbors/source flux` for a
```
taranu: "This is" can be removed. I would specify "within the source's footprint", if that's what it is, and also move the justification down to the Notes.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -452,6 +452,97 @@ def deblend_lite(mExposure, footprint, config, wavelets=None):
...
+
+    def __str__(self):
+        """Show all of the metrics in the class
+        """
+        string = (f"DeblenderMetrics<"
+                  f"maxOverlap={self.maxOverlap}"
+                  f"fluxOverlap={self.fluxOverlap}"
+                  f"overlapRatio={self.overlapRatio}"
+                  f"blendedness={self.blendedness}>")
+        return string
```
taranu: Whether you keep the array structure or not, this really should be a Dataclass. You get `__init__` and `__str__` (well, probably `__repr__`) for free. Plus you can make it frozen, which admittedly wouldn't help with changing individual values in an array but would at least prevent reassignment of the field. E.g.:
```@dataclass(frozen=True)
class DeblenderMetrics:
    maxOverlap: numpy.ndarray
    fluxOverlap: numpy.ndarray
```

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -452,6 +452,97 @@ def deblend_lite(mExposure, footprint, config, wavelets=None):
...
+
+
+def setDeblenderMetrics(blend):
+    """Set metrics that can be used to evalute the deblender accuracy
+
+    Without a full Bayesian analysis it is difficult to evaluate the
```
taranu: The rest of this docstring doesn't really seem to add anything to the short summary. Since the metrics are defined in the `DeblenderMetrics` class anyway, I think you can just say that this method computes the `DeblenderMetrics` for each `source` in the `blend` and assigns it to the `metric` property.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -452,6 +452,97 @@ def deblend_lite(mExposure, footprint, config, wavelets=None):
...
+    ----------
+    blend : `scarlet.lite.Blend`
+        The blend containing the sources to measure.
+    """
+    # Store the full model of the scene for comparison
+    blendModel = blend.get_model()
+    for k, src in enumerate(blend.sources):
+        # Extract the source model in the full bounding box
+        model = src.get_model(bbox=blend.bbox)
+        # The footprint is the 2D array of non-zero pixels in each band
```
taranu, line 505: Is this a standard Scarlet definition? It seems pretty important. I suppose with monotonicity on it's guaranteed to be a contiguous region but is that generally true?

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -452,6 +452,97 @@ def deblend_lite(mExposure, footprint, config, wavelets=None):
...
+        footprint = np.bitwise_or.reduce(model > 0, axis=0)
```
taranu, line 506: I presume that's the band axis? Are there any `NaN`s around?

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -452,6 +452,97 @@ def deblend_lite(mExposure, footprint, config, wavelets=None):
...
+        # Calculate the metrics.
+        # See `DeblenderMetrics` for a description of each metric.
+        neighborOverlap = (blendModel-model) * footprint[None, :, :]
```
taranu, line 509: I suppose this must be strictly positive based on the definition of the footprint, presuming that `blendModel`>`model` everywhere?

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -740,21 +831,32 @@ def __init__(self, schema, peakSchema=None, **kwargs):
...
+                                             "This includes peaks that may have been culled "
+                                             "during deblending or failed to deblend")
+        # Skipped flags
+        self.deblendSkippedKey = schema.addField('deblend_skipped', type='Flag',
+                                                 doc="Deblender skipped this source")
+        self.isolatedParentKey = schema.addField('deblend_isolatedParent', type='Flag',
+                                                 doc='The source has only a single peak '
+                                                     'and was not deblended')
+        self.pseudoKey = schema.addField('deblend_isPseudo', type='Flag',
+                                         doc='The source is identified as a "pseudo" source and '
```
taranu, line 824: Non-detection (e.g. skyObject)?

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -452,6 +452,97 @@ def deblend_lite(mExposure, footprint, config, wavelets=None):
...
+    blendModel = blend.get_model()
+    for k, src in enumerate(blend.sources):
+        # Extract the source model in the full bounding box
+        model = src.get_model(bbox=blend.bbox)
+        # The footprint is the 2D array of non-zero pixels in each band
+        footprint = np.bitwise_or.reduce(model > 0, axis=0)
+        # Calculate the metrics.
+        # See `DeblenderMetrics` for a description of each metric.
+        neighborOverlap = (blendModel-model) * footprint[None, :, :]
+        maxOverlap = np.max(neighborOverlap, axis=(1, 2))
```
taranu, line 510: This is the maximum value in a single pixel, correct? The docs don't really make that clear.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -798,16 +894,45 @@ def _addSchemaKeys(self, schema):
...
+                                              doc="Number of components in a ScarletLiteSource. "
+                                                  "If `config.version != 'lite'`then "
+                                                  "this column is set to zero.")
+        self.psfKey = schema.addField('deblend_deblendedAsPsf', type='Flag',
+                                      doc='Deblender thought this source looked like a PSF')
+        # Blendedness/classification metrics
+        self.maxOverlapKey = schema.addField("deblend_maxOverlap", type=np.float32,
+                                             doc="Maximum overlap with all of the other neighbors flux "
+                                                 "combined."
+                                                 "This is useful as a metric for determining how blended a "
```
taranu, line 889: I don't think we are supposed to put extended justifications or usage suggestions like this into a field doc - just define the quantity.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -798,16 +894,45 @@ def _addSchemaKeys(self, schema):
...
+                                                 "at or below the noise level, it is likely to be a mostly "
+                                                 "isolated source in the deconvolved model frame.")
+        self.fluxOverlapKey = schema.addField("deblend_fluxOverlap", type=np.float32,
+                                              doc="This is the total flux from neighboring objects that "
+                                                  "overlaps with this source.")
+        self.overlapRatioKey = schema.addField("deblend_fluxRatio", type=np.float32,
+                                               doc="This is the fraction of "
+                                                   "`flux from neighbors/source flux` "
+                                                   "for a given source. "
+                                                   "This is potentially more useful than the canonical "
```
taranu: Remove this sentence and briefly specify e.g. "within the source's footprint" instead.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -798,16 +894,45 @@ def _addSchemaKeys(self, schema):
...
+                                                   "'blendedness' (or purity) metric because it accounts "
+                                                   "for potential biases created during deblending by not "
+                                                   "weighting the overlapping flux with the flux of this "
+                                                   "sources model.")
+        self.blendednessKey = schema.addField("deblend_blendedness", type=np.float32,
+                                              doc="The metric for determining how blended a source is "
```
taranu: Could just be "The Bosch et al. 2018 metric for 'blendedness.'"

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -1123,6 +1217,19 @@ def deblend(self, mExposure, catalog):
                       nParents, nDeblendedParents, nChildren, len(catalog))
         return catalogs, weightedCatalogs
 
+    def _setMetrics(self, source, metrics, filterIndex):
+        """Set the metrics for a source in either a template or weighted
+        catalog.
+
+        Parameters
+        ----------
+
```
taranu: Missing docstrings.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -1196,7 +1321,81 @@ def _skipParent(self, parent, skipKey, logMessage):
         parent.set(self.deblendSkippedKey, True)
         parent.set(skipKey, True)
 
-    def _updateParentRecord(self, parent, nPeaks, nChild,
+    def _isSkipped(self, parent, mExposure):
```
taranu: If the purpose of this function is to update the record, then I think the name should stay `_updateSomething` (`_updateSkipped`?). I would not expect a function named `isSomething` to modify anything.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -1196,7 +1321,81 @@ def _skipParent(self, parent, skipKey, logMessage):
...
+            `True` if the deblender will skip the parent
+        """
+        skipKey = None
+        skipMessage = None
+        footprint = parent.getFootprint()
+        if len(footprint.peaks) < 2 and not self.config.processSingles:
+            # Skip isolated sources unless processSingles is turned on.
+            # Note: this does not flag isolated sources as skipped or
+            # set the NOT_DEBLENDED mask in the exposure,
+            # since these aren't really a skipped blends.
```
taranu: Remove "a"

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -452,6 +452,97 @@ def deblend_lite(mExposure, footprint, config, wavelets=None):
...
+        # Calculate the metrics.
+        # See `DeblenderMetrics` for a description of each metric.
+        neighborOverlap = (blendModel-model) * footprint[None, :, :]
+        maxOverlap = np.max(neighborOverlap, axis=(1, 2))
+        fluxOverlap = np.sum(neighborOverlap, axis=(1, 2))
+        fluxModel = np.sum(model, axis=(1, 2))
+        fluxOverlapFraction = np.zeros((len(model), ), dtype=float)
+        isFinite = fluxModel > 0
+        fluxOverlapFraction[isFinite] = fluxOverlap[isFinite]/fluxModel[isFinite]
+        blendedness = 1 - np.sum(model * model, axis=(1, 2)) / np.sum(blendModel * model, axis=(1, 2))
```
taranu: Unnecessary spaces around `*` and `/`.

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -148,11 +149,11 @@ def modelToHeavy(source, mExposure, blend, xy0=Point2I(), dtype=np.float32):
     return mHeavy
 
 
-def liteModelToHeavy(source, mExposure, blend, xy0=None, dtype=np.float32, useFlux=False):
+def liteModelToHeavy(source, mExposure, blend, xy0=Point2I(), dtype=np.float32, useFlux=False):
     """Convert a scarlet model to a `MultibandFootprint`.
     Parameters
     ----------
-    source : `scarlet.LiteSource`
+    source : `scarlet.Source`
```
taranu: Did `LiteSource` get folded back into `Source`?

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -131,10 +131,11 @@ def modelToHeavy(source, mExposure, blend, xy0=Point2I(), dtype=np.float32):
     # Always use a real space convolution to limit artifacts
     model = blend.observations[0].renderer.convolve(model, convolution_type="real").astype(dtype)
     # Update xy0 with the origin of the sources box
-    xy0 = Point2I(overlap.origin[-1] + xy0.x, overlap.origin[-2] + xy0.y)
+    # Update xy0 with the origin of the sources box
```
taranu: Duplicate comment

python/lsst/meas/extensions/scarlet/source.py
```diff
@@ -210,15 +207,16 @@ def liteModelToHeavy(source, mExposure, blend, xy0=None, dtype=np.float32, useFl
         model = blend.observation.convolve(model, mode="real").astype(dtype)
 
     # Update xy0 with the origin of the sources box
-    xy0 = Point2I(overlap.origin[-1] + xy0.x, overlap.origin[-2] + xy0.y)
+    _xy0 = Point2I(overlap.origin[-1] + xy0.x, overlap.origin[-2] + xy0.y)
```
taranu: I still think you should just change the function args to separate x0, y0 ints if you don't actually need the `Point2I` object. Having both `xy0` and `_xy0` in scope is bound to confuse someone eventually.

taranu (commented):
Mostly minor nitpicks, but hopefully more useful than annoying.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: I liked your `dataclass` suggestion and implemented it.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: I didn't know about this. Thanks for the suggestion!

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: There shouldn't be. Any NaN's in the model would produce NaN's in the gradient and cause a numpy error. So there are explicit checks to make sure that this doesn't happen.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: Every source that we use in the science pipelines is monotonic. Even if that changes in the future, the footprint is still the non-zero pixels (the same way that a footprint is defined in `afw`). All models also have a positivity constraint. If we start allowing sources to have negative flux then numerous changes would be required across the task (and internal scarlet lite code).

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: Yep, unless something changes in the future and we start allowing models to have negative flux. But again, that would create a host of other problems and since it's unlikely that we'll implement that, I think that it would be overkill to try to make things more general right now.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: Yes, but the `SetPrimaryFlagsTask` technically allows for other types of non-sources, so since it doesn't cost us anything we use that same definition here. This was just a schema key that was reordered, not changed in any way.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: Really, I thought that we were? I've always done that in my field docs.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: This is just git being weird. The `_updateParentRecord` method is still there and was refactored to somewhere else, but for some reason git did the diff this way.

python/lsst/meas/extensions/scarlet/source.py
fred3m: No, good catch.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
timj: Where would we put the usage suggestion? Is there guidance on this in the dev guide?

python/lsst/meas/extensions/scarlet/source.py
fred3m: Typically there is an `xy0` (see [`ScarletDeblendTask`](https://github.com/lsst/meas_extensions_scarlet/blob/main/python/lsst/meas/extensions/scarlet/scarletDeblendTask.py#L1025)), but it also needs to be adjusted by the location of the minimum bounding box that contains the footprint. So there has to be a second `xy0`. I guess I could just overwrite the initial `xy0` since it's never used again,.

fred3m: I think at one point I was testing something and needed them separate, but forgot to switch it back.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
taranu: I don't think so, and said as much on Jira. I guess I'm fine with a brief explanatory comment like "values close to the noise level indicate minimal blending". I assume most of the columns that end up in `objectTable`s are defined in LPMs or similar docs. If usage suggestions are based on applications to specific datasets like test-med-1/RC2, those should probably go in a DMTN.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
taranu: I realize that, but this function also updates the parent by potentially calling `_skipParent`.

fred3m: I see. Would `_checkSkipConditions` be more suitable? The point of the function really is to check whether or not a source should be skipped, it's only updated in that method because it's easier to update the parent record while it has all of the information about why it is being skipped.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: I agree that we do need deblender docs and it's on me for not having a good DMTN or page(s) in the pipeline docs to discuss how meas_extensions_scarlet works and how its outputs should be used. I've just put off making them because I want things to feel more finished. I know that's a bar that can easily continue to keep moving but I feel like the current blockers are a fair place to start. See DM-30535, which I just updated with an appropriate set of blockers.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
taranu: I think `_check` also doesn't imply any updates. I see that you use this function in both `setSkipFlags`, where it's only called for the side effect without checking the return value, and elsewhere only to check the return value. So if you want to keep it one function, it should reflect the fact that it can change `parent` if necessary. Otherwise, I suppose you could have it do something like return either `None` or `(skipKey, skipMessage)` and not do the modifications itself.

fred3m: I used the second option and kept it in its own commit. This gave me my first opportunity to use the walrus operator, but I'll wait for your approval before re-running Jenkins and merging.

taranu: That seems like it should work. I generally prefer the explicit `(skipArgs := self._checkSkipped(src, mExposure)) is None` to the version without `is None` (you have one instance of each).

fred3m: Yep, that was me being careless. Thanks for catching it.

PR #53, opened 2022-05-18
DM-33710: store scarlet models in place of footprints

fred3m:


python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -0,0 +1,808 @@
...
+    xy0 : `tuple` of `int`
+        The lower bound of the components bounding box.
+    extent : `tuple` of `int`
+        The `(width, height)` of the component array.
+    center : `tuple` of `int`
+        The center of the component.
+    model : `numpy.ndarray`
+        The model for the component.
+    """
+    xy0: tuple[int]
```
TallJimbo: It's not being checked by anything, but this annotation actually means "tuple with one int element"; you want `tuple[int, int]` throughout this file.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -0,0 +1,808 @@
...
+        ----------
+        catalog : `lsst.afw.table.SourceCatalog`
+            The catalog missing heavy footprints for deblended sources.
+        band : `str`
+            The name of the band that the catalog data describes.
+        psfModel : `lsst.afw.detection.Psf`
+            The observed PSF model for the catalog.
+            If `psfModel` is `None` and `redistributeCoadd` is `None`
+            then the models are not convolved.
+        redistributeCoadd : `lsst.afw.image.Exposure`
```
TallJimbo: While we're only running this on coadds, I don' think _this_ code cares about it being a coadd.  Maybe rename to `redistributeImage`?  I'd also endorse making it just an `Image`, and requiring `psfModel` to always be passed separately, if those those are actually the only components you need.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -0,0 +1,808 @@
...
+        # Get the index of the model for the given band
+        bandIndex = self.filters.index(band)
+
+        for parentRecord in parents:
+            parentId = parentRecord.getId()
+
+            try:
+                blendModel = self.blends[parentId]
+            except KeyError:
+                warnings.warn(f"Unable to load models found for parent source {parentId}")
```
TallJimbo: Do you have a specific failure mode in mind?  This seems like something that should either be an exception or not a warning.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -0,0 +1,808 @@
...
+
+    Attributes
+    ----------
+    xy0 : `tuple` of `int`
+        The lower bound of the components bounding box.
+    extent : `tuple` of `int`
+        The `(width, height)` of the component array.
+    center : `tuple` of `int`
+        The center of the component.
+    model : `numpy.ndarray`
```
TallJimbo, line 36: What happens when the built-in `json` module sees a numpy array?  I can think of two possibilities, neither of which sounds great:

- some kind of portable binary blob encoded as base64 or hex -> inefficient storage unless we also gzip or similar (and I don't think we are given the changes in `daf_butler`);
- transformation into a JSON list of numbers with strides (or a nested list of numbers) that are saved as strings -> inefficient to read, and either slightly lossy or even more inefficient without compression.

I know you've been consulting with @natelust on this, and maybe he'd say this is not as much of a problem as I'm imagining (and he'd know better than I would).  But I'm pretty surprised to see JSON used for floating-point array storage rather than a native-binary format.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -0,0 +1,808 @@
...
+    """Create footprints and update band-dependent columns in the catalog
+
+    Parameters
+    ----------
+    blendData : `ScarletBlendData`
+        Persistable data for the entire blend.
+    catalog : `lsst.afw.table.SourceCatalog`
+        The catalog that is being updated.
+    modelPsf : `numpy.ndarray`
+        The 2D model of the model PSF.
```
TallJimbo:
```suggestion
        The 2D model of the PSF.
```
Or is "model" meaning two different things here?

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -0,0 +1,808 @@
...
+    observedPsf : `lsst.afw.detection.Psf`
+        The observed PSF model for the catalog.
+        If `observedPsf` is `None` and `redistributeCoadd` is `None`
+        then the models are not convolved.
```
TallJimbo: Is not convolving functionality we might actually use?  Or is it extra complexity that's most likely to be used in error?

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -0,0 +1,808 @@
...
+        then the PSF of the `resdistributeCoadd` is used.
+    bandIndex : `int`
+        The number of the band to extract.
+    parentFootprint : `lsst.afw.Footprint`
+        The footprint of the parent, used for masking out the model
+        when re-distributing flux.
+    updateFluxColumns : `bool`
+            Whether or not to update the `deblend_*` columns in the catalog.
+            This should only be true when the input catalog schema already
+            contains those columns.
```
TallJimbo: Indentation.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -0,0 +1,808 @@
...
+            blend=blend,
+            xy0=xy0,
+            useFlux=useFlux,
+        )
+        sourceRecord.setFootprint(heavy)
+
+        if updateFluxColumns:
+            # Set the flux of the scarlet model
+            # TODO: this field should probably be deprecated,
+            # since DM-33710 gives users access to the scarlet models.
```
TallJimbo, line 488: Time to write the ticket, if you haven't already.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -0,0 +1,808 @@
...
+
+    def get_model(self, bbox=None):
+        """Generate the model for the source
+
+        Parameters
+        ----------
+        bbox : `scarlet.bbox.Box`
+            The bounding box to insert the model into.
+            If `bbox` is `None` then the model is returned in its own
+            bounding box.
```
TallJimbo, line 564: Missing `Returns` docs.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -0,0 +1,808 @@
...
+        Persistable data for the entire blend.
+    nBands : `int`
+        The number of bands in the image.
+        If `bandIndex` is `None` then this parameter is ignored and
+        the number of bands is set to 1.
+    bandIndex : `int`
+        Index of model to extract. If `bandIndex` is `None` then the
+        full model is extracted.
+    dtype : `numpy.dtype`
+        The data type of the model that is generated.
```
TallJimbo, line 650: Missing `Returns` docs.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -0,0 +1,808 @@
...
+@dataclass
+class ScarletComponentData:
+    """Data for a component expressed as a 3D data cube
+
+    For now this is used for scarlet main source models because
+    their structure is too complex to persist in the same
+    way that scarlet lite components can be persisted.
+
+    Note that both `xy0` and `extent` use lsst ``(x, y)`` convention,
+    not the scarlet/C++ ``(y, x)`` convention.
```
TallJimbo, line 26: Don't feel compelled to use LSST conventions for these if they make things more difficult, as it seems they might.  I actually tend to think of the boundary as not being LSST vs. numpy but rather "`Point` and `Extent` use (x, y)" while tuples use `(y, x)`", since `Image.array` is actually also indexed as `[y, x]`.  (There is an exception to that, in that `lsst.skymap` uses `(x, y)` tuples for patch indexes, but at least we recently made that a `NamedTuple` as a partial mitigatoin).

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -0,0 +1,808 @@
...
+    for source in blend.sources:
+        componentData = ScarletComponentData(
+            xy0=tuple(int(x) for x in source.bbox.origin[1:][::-1]),
+            extent=tuple(int(x) for x in source.bbox.shape[1:][::-1]),
+            center=tuple(int(x) for x in source.center[::-1]),
+            model=source.get_model(),
+        )
+
+        sourceData = ScarletSourceData(
+            modelType=source.__class__.__name__,
```
TallJimbo: Using `lsst.utils.introspection.get_full_type_name` here and `lsst.utils.doImportType` on the other end might add some robustness, though it will also encode the full path to the class (making it hard to move things around later).  If you want to limit this to an limited set of types instead, you might want to define an `Enum`, perhaps with `type` values, to enforce that at save-time so nobody is sad days later at load-time that they saved an unsupported model type.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -125,16 +123,15 @@ def _computePsfImage(self, position):
             # data in some bands, so we track this error to distinguish it
             # from unknown errors.
             msg = "Failed to compute PSF at {} in band {}"
-            raise IncompleteDataError(msg.format(position, self.filters[bidx]))
+            raise IncompleteDataError(msg.format(position, filters[bidx]))
```
TallJimbo: Add `from None` (if you want the traceback to end here) or `from err` (with `as err` in the `except` earlier) to avoid extra "while handling this exception...another exception" noise in the traceback.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -424,22 +419,23 @@ def deblend_lite(mExposure, footprint, config, wavelets=None):
...
-    blend.fit(max_iter=config.maxIter, e_rel=config.relativeError, min_iter=config.minIter)
+    blend.fit(
+        max_iter=config.maxIter,
+        e_rel=config.relativeError,
+        min_iter=config.minIter,
+        reweight=False,
+    )
 
-    return blend, skipped, spectrumInit
+    # Convert the blend into a ScarletBlendData instance
```
TallJimbo: This comment doesn't seem right.

tests/test_deblend.py
```diff
@@ -101,81 +105,133 @@ def test_deblend_task(self):
...
-            denseFootprint[3] = True
-            np.testing.assert_array_equal(denseFootprint, _catalog["deblend_tooManyPeaks"])
-
-            # Check that only the appropriate parents were skipped
-            skipped = largeFootprint | denseFootprint
-            np.testing.assert_array_equal(skipped, _catalog["deblend_skipped"])
+        catalog, dataModel = deblendTask.run(coadds, catalog)
+
+        # Attach the footprints in each band and compare to the full
+        # data model. This is done in each band, bth with and without
```
TallJimbo:
```suggestion
        # data model. This is done in each band, both with and without
```

TallJimbo (approved):

python/lsst/meas/extensions/scarlet/io.py
fred3m: Sure, that makes sense.

python/lsst/meas/extensions/scarlet/io.py
fred3m: Yes. If the parent blend was skipped then there isn't an entry for it in the models. I mostly had this for testing, since I wanted to make sure that every parent that wasn't skipped had a model. I could keep a list of skipped parents, but that seems like a waste of data, so I'm inclined to just `pass` n a `KeyError`. Thoughts?

python/lsst/meas/extensions/scarlet/io.py
fred3m: @natelust can comment better than me. It is true that it is storing the models in json as strings, which doesn't seem very efficient, but preliminary testing shows that they still seem to take up less data than having HeavyFootprints in each band. In the long run I agree that we should probably have some sort of hierarchical binary format, but Nate seemed to think that there were reasons we couldn't do that.

python/lsst/meas/extensions/scarlet/io.py
fred3m: That's a good point. While storing the deconvolved models is useful, and may be useful for users downstream, I don't ever see us adding those to a _catalog_. I'll make this a mandatory argument.

python/lsst/meas/extensions/scarlet/io.py
fred3m: Does this require an RFC?

python/lsst/meas/extensions/scarlet/io.py
fred3m: Thanks, but I think that in this case it makes sense because `xy0` (not `yxo`) and `extent` are LSST definitions and the `center` is ` Point2I`, so it makes sense to use that convention. Even if I used the opposite, I would still contain a line in the docstring to avoid potential confusion.

python/lsst/meas/extensions/scarlet/io.py
fred3m: Ah, so fortunately the name is just meta data and not actually used in reconstruction. In scarlet lite there is (and should always be) only one type of source class, whereas that was a much more important thing to know in scarlet main. Since this is already stored in the `deblendedCatalg` (https://github.com/lsst/meas_extensions_scarlet/blob/main/python/lsst/meas/extensions/scarlet/scarletDeblendTask.py#L1450) I can just remove this altogether.

python/lsst/meas/extensions/scarlet/io.py
fred3m: In fact, the tests make sure that all skipped parents don't show up here, and that all deblended parents do, so I do think that the warning is unnecessary.

python/lsst/meas/extensions/scarlet/io.py
TallJimbo: If it's coming from a config field, yes, that would need an RFC.  If it's just some function argument, then I think probably not unless it's prominent in some task.  Boundary between public and [package-]private interfaces here is a little blurry.

TallJimbo (approved):

PR #57, opened 2022-07-12
DM-35530: Fix typo in persisted scarlet model data

fred3m:


python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -130,7 +130,7 @@ def asDict(self) -> dict:
         return {
             "xy0": self.xy0,
             "extent": self.extent,
-            "center": self.extent,
+            "center": self.center,
```
laurenam, line 135: Because of the mix-n’-match of conventions, it wouldn’t hurt to be explicit about (y, x) vs. (x, y) in the variable doc strings (in addition to your comment in the class description above).  I got nervous looking at your test…!

laurenam (approved):

python/lsst/meas/extensions/scarlet/io.py
fred3m: Good point, I'll make it clear in the docs.

fred3m: It's actually more confusing than you think. To be consistent with xy0 and the rest of the stack, the center in the persistence model is in `(x, y)` format, it's just `scarlet` that uses the C++ `(y, x)` format. But I am clarifying in the docs.

laurenam: 🤪

PR #58, opened 2022-07-12
DM-30535: Update the docs, including API changes

fred3m:


doc/lsst.meas.extensions.scarlet/models.rst
```diff
@@ -0,0 +1,260 @@
...
+.. code-block:: python
+
+    import lsst.meas.extensions.scarlet as mes
+
+    # Select the first record with exactly 5 deblended children
+    parent = catalog[catalog["deblend_nChild"]== 5][0]
+    # Load the PSF for the observation in each band
+    observedPsfs = []
+    for band in modelData.bands:
+        observedPsfs.append(butler.get("deepCoadd_calexp.psf", tract=tract, patch=patch, band=band))
```
taranu: Why not just make this a list comprehension?

doc/lsst.meas.extensions.scarlet/models.rst
```diff
@@ -0,0 +1,260 @@
...
+    blend = mes.io.multibandDataToScarlet(
+        modelData=modelData,
+        blendId=parent.getId(),
+        mExposure=mExposure,
+        footprint=parent.getFootprint()
+    )
+
+.. note::
+    When using a :py:class:`~lsst.afw.image.multiband.MultibandExposure` it is not necessary to also load the PSF, since the :py:class:`~lsst.afw.image.multiband.MultibandExposure` already contains the PSF model.
+    Also note that we incldued the ``footprint`` parameter, which will mask out all regions in the exposure outside of the parent footprint.
```
taranu: included

doc/lsst.meas.extensions.scarlet/models.rst
```diff
@@ -0,0 +1,260 @@
...
+    # Use the Lupton RGB sinh^-1 mapping to preserve colors
+    norm = scarlet.display.AsinhMapping(minimum=0, stretch=0.1, Q=10)
+    # Convolve the model into the observed seeing and discard the narrow-band filter
+    model = blend.get_model(convolve=True)[1:]
+    # Create a mask to hide the footprint
+    mask = ~parent.getFootprint().spans.asArray()
+    # Convert the multiband model into 3 RGB colors
+    rgb = scarlet.display.img_to_rgb(model, norm=norm, mask=mask)
+    # Display the model
+    plt.imshow(rgb)
```
taranu: Consider `origin='lower'` here instead of `invert_yaxis()` below.

doc/lsst.meas.extensions.scarlet/models.rst
```diff
@@ -0,0 +1,260 @@
...
+        norm=norm,
+        channel_map=channelMap,
+        show_model=True,
+        show_rendered=True,
+        show_observed=True,
+        show_residual=True,
+    )
+    plt.show()
+
+.. image:: images/reconstructedBlendDisplay.png
```
taranu, line 158: Can you add an option to show_scene to plot a 2x2 grid instead of 4x1? The docs are fairly narrow and having four side-by-side makes them tiny. Actually, I see you had a 2x2 grid for the images at the end of the Overview so maybe this already exists.

In that case I'd suggest having Model Rendered and Observation on the top row. Consider renaming Model Rendered to Convolved Model as it took me a few moments to realize that's what it meant.

doc/lsst.meas.extensions.scarlet/models.rst
```diff
@@ -0,0 +1,260 @@
...
+        show_model=True,
+        show_rendered=True,
+        show_observed=True,
+        show_residual=True,
+    )
+    plt.show()
+
+.. image:: images/extraSrcBlend.png
+
+Looking at source 3, and the residual image, we can now see why this source isn't modeled properly by scarlet.
```
taranu, line 261: ... source 3? I'm confused. Comparing with the previous image, the fit for it did improve, but I thought we were focusing on source 0?

For that matter, flipping between the two images, it's unclear why source 2 improved so much, or rather why it wasn't modelled well originally. Is it because 2-3-0-11 are all in a line? But it's also not clear why source 0 originally had a thin diagonal line. It just looks like the original fit didn't converge well.

Anyway, I don't think you need to do a complete breakdown of exactly what happened in every step of this example, but since you are showcasing it, some clarification would help.

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+
+.. image:: images/template0.png
+
+Even in this case the template isn't perfect (and isn't intended to be), which we see by looking at the residual of the image with the source 0 template subtracted:
+
+.. image:: images/templateResidual0.png
+
+Most of the residual that is part of the actual source is due to the non-symmetric shape of the PSF, however there is also a small blob to the right due to an undetected source that will eventually be falsy included in the flux of this source when we re-apportion the flux in the image (see `re-apportion`_).
+
+Looking at source 2 we see another artifact caused by a complicated galaxy that has multiple non-monotonic regions that either need to be modeled as additional components or a more complex non-monotonic model (see :ref:`scarlet_restart`).
```
taranu, line 47: This particular image makes it look like a close pair of sources rather than one complicated source.

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+.. image:: images/template3.png
+
+Here, the template for the fainter source in the center picks up flux from both of its neighbors, which will cause it to steal flux from both of them and affect all three models once we re-apportion the flux.
+Attempts have been made to use a monotonic projection for templates to mitigate this problem but those tend to cause large fractal-like structures that have been determined to make the templates worse in general, so they are not used in production (but using them does exist via a config option).
+
+PSF Templates
+^^^^^^^^^^^^^
+
+Sources with very low signal to noise (like all sources labeled 5 and greater in the above blend) are unlikely to yield very good symmetric templates because the models are dominated by noise and quickly truncate.
+So :py:mod:`~lsst.meas.deblender` attempts to fit all sources to the PSF (including fitting the centroid and a linear background model).
```
taranu: fit the PSF (or a PSF model) to all sources

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+.. image:: images/template2.png
+
+These, and other types of "unrecognized blends" are unavoidable and can be expected in all ground based surveys, including HSC and LSST.
+
+Finally we have source 3, which shows the worst deblending case for a symmetry based deblender: three sources that are nearly aligned and equally spaced
+
+.. image:: images/template3.png
+
+Here, the template for the fainter source in the center picks up flux from both of its neighbors, which will cause it to steal flux from both of them and affect all three models once we re-apportion the flux.
+Attempts have been made to use a monotonic projection for templates to mitigate this problem but those tend to cause large fractal-like structures that have been determined to make the templates worse in general, so they are not used in production (but using them does exist via a config option).
```
taranu: Which config option?

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+
+PSF Templates
+^^^^^^^^^^^^^
+
+Sources with very low signal to noise (like all sources labeled 5 and greater in the above blend) are unlikely to yield very good symmetric templates because the models are dominated by noise and quickly truncate.
+So :py:mod:`~lsst.meas.deblender` attempts to fit all sources to the PSF (including fitting the centroid and a linear background model).
+Any sources with a sufficiently small residual (determined by the task config) will use the PSF template in place of the symmetric template.
```
taranu: Which config option? (again)

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+
+.. note::
+    Unfortunately the PSF model is not good enough to fit bright sources, even if they are perfectly isolated.
+    So typically the only sources that use the PSF template as a model are the very faintest sources.
```
taranu, line 70: But the deeper the coadd, the less likely a faint source is to actually be a star (or quasar). And even the faintest galaxies still have measurable finite sizes.

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+
+PSF Templates
+^^^^^^^^^^^^^
+
+Sources with very low signal to noise (like all sources labeled 5 and greater in the above blend) are unlikely to yield very good symmetric templates because the models are dominated by noise and quickly truncate.
+So :py:mod:`~lsst.meas.deblender` attempts to fit all sources to the PSF (including fitting the centroid and a linear background model).
+Any sources with a sufficiently small residual (determined by the task config) will use the PSF template in place of the symmetric template.
+
+.. note::
+    Unfortunately the PSF model is not good enough to fit bright sources, even if they are perfectly isolated.
```
taranu, line 68: It might be worth expanding on this, possibly with an example. Or at least comment on whether the PSF model matters (PSFEx vs Piff).

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+
+.. note::
+    Source 3 was subtracted using the model obtained in `multi-band_deblending`_.
+
+But even this model can be improved.
+While source 0 is all of the flux in its footprint (since no other model overlaps with it), there is still some mixing between sources 1 and 2.
+This is partly due to one or two undetected sources in source 2, but also because symmetric templates alone are not enough to separate objects in a monochromatic image.
+A monotonic projection can help, but looking back at the full color image we see that the easiest way for our eyes to separate the sources is by using color information, which will be discussed in the next section.
+
+See `Lupton 2001 <https://www.astro.princeton.edu/~rhl/photomisc/deblender.pdf>`_ for more details on the SDSS implementation of this algorithm.
```
taranu, line 114: Is there no actual reference for this? At least a DOI or something?

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+
+.. _multi-band_deblending:
+
+Multi-band Deblending with scarlet (lite)
+-----------------------------------------
+
+The difference between scarlet and scarlet lite
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The science pipelines currently use models generated by the "lite" version of *scarlet*, so it's worth understanding the difference between the two versions.
```
taranu, line 124: Is this the best place for this, though? I'm not seeing a lot that's directly relevant for science pipelines users, especially if the pipelines are going to exclusively use scarlet lite. Maybe a link to scarlet lite's own documentation would suffice.

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+However, the underlying models and algorithms in scarlet and scarlet-lite are *identical*, the only difference between them is the architecture of the code used to generate and optimize the models.
+For this reason the remainder of this document will refer to the multi-band deblending algorithm as *scarlet*, even though in practice the actual package implementing the algorithm is *scarlet lite*.
+
+.. _basic_scarlet_model:
+
+Basic Model
+^^^^^^^^^^^
+
+Just like the single-band deblender, in order to make a model for a multi-band image we will have to make some assumptions.
+In this case the main assumption is that an astrophysical source can be modeled as a collection of components, where each component can be described by a single SED that is roughly constant but can vary in magnitude over its spatial extent (morphology).
```
taranu: Define SED (the acronym, at least).
I don't think broadband fluxes qualify as an SED, not in the least because multiple (or really, an infinite selection of) SEDs can produce identical broadband fluxes.
If the Python class is called Spectrum I would define its meaning here and use that term thereafter.

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+This assumption can be justified physically by considering the structure of a galaxy.
+In general the populations of stars in the bulge and disk are different, as well as perhaps additional populations of star forming regions, each with its own distinct SED.
+However, the actual colors in each population are not that different, which is why parametric bulge-disk models (Sersic models, Spergel models, Gaussian mixture models, etc.) have been successfully used in deblending and photometric measurement.
```
taranu: I'm not really sure what you mean by "the actual colors in each population are not that different".

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+
+The other assumption that is currently made in the science pipelines models is that each component is monotonically decreasing from a peak.
+Again, thinking of galactic structure this is usually a reasonable assumption as both the bulge and disk are brightest in the center, and even spiral arms generally emanate from a peak location.
+There are exceptions, however, like nebula and irregular galaxies that are much less organized and not well constrained by a component-like model with a center.
```
taranu: nebulae; "less organized" vs "asymmetric"?

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+
+The other main astrophysical effect that is overlooked by these assumptions is extinction by gas and dust, which can cause reddening over a single component.
+While it is outside of the discussion in this tutorial, we have observed reddening residuals in edge-on galaxies and it is currently an active area of research to better model galactic (and intergalactic) dust using *scarlet*.
```
taranu: outside of the scope of this tutorial

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+
+Because the *scarlet* model is multi-band, and the observations are not required (or recommended) to be matched to a common PSF, the model itself must exist in a space that can be convolved into the observed seeing in each band, which we will refer to as *model space*.
+In order to be sufficiently well sampled we currently use a PSF for the model space that is a circular Gaussian with ``σ=0.8`` pixels, ensuring that our model image is well-sampled but smaller than the observed PSF in each observed band.
```
taranu, line 162: The reasons for this is so that you can use pixel-space (FFT) convolution at the native pixel scale without oversampling, yes? It's probably worth highlighting that.

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+    However, there are currently no plans to use *scarlet* on single visits and it is expected that co-adds will always be well-sampled.
+
+.. _model_init:
+
+Initialization
+^^^^^^^^^^^^^^
+
+One way to think of the initial *scarlet* models is that they are basically the same as the templates generated by the single band deblender (see `single-band_deblending`_) with one key difference: because the *scarlet* models are obtained using an iterative optimization procedure, using the assumption of monotonicity from the peak location is a well-defined and useful (perhaps necessary) constraint.
+
+For *scarlet* we have chosen the *𝛘²* (inverse variance-weighted or Kaiser) co-add to initialize our morphologies.
```
taranu: Briefly explain this, and elaborate on whether this a dataset in the butler, or something you generate on-the-fly?

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+While this is not a truly accurate model (since each observation uses a different seeing), it gives us a fast and efficient method for generating a detection image.
+The main downfall of this choice is that our model space has a much narrower seeing than the detection image, meaning our initial models will always be "puffy."
+We have tried numerous deconvolution and approximation methods but so far we have not found any method that provides enough of an improvement to overcome the runtime required to generate the detection image.
+In other words, every method that we have come up with to generate a better detection image is more expensive computationally than just running the scarlet optimizer for a few extra iterations to converge to the same solution.
+
+Once we have fit the morphology for each component of each source, we use a least-squares algorithm to simultaneously fit the SED of each source *in scarlet model space*.
```
taranu: Linear least-squares, presumably?

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+Because we normalize the morphology, including the intensity information in the SED, this is different than the observed SED, as the PSF in each band will distribute the model flux differently.
+In this case the computational savings of the least-squares SED fit *is* significant enough to justify its use.
+
+.. note::
+    For *very* large (approx *1k* × *1k* pixel) blends the least squares fit runs into memory issues, since the size of the array to fit is roughly ``number of sources × number of bands × width × height)``.
+    In that case setting :attr:`~lsst.meas.extensions.scarlet.ScarletDeblendConfig.setSpectra` to ``False`` is recommended.
+
+.. note::
+    An interesting fact about optimization in scarlet is that the best initialization (in terms of 𝛘²) will not necessarily result in the fastest runtime or even the most optimal fit.
```
taranu, line 191: Isn't this a general result of being trapped in a local minimum?

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+    The interplay between the particular gradient descent method used and proximal operators is complicated and often certain irregularities (like sources with slightly worse positions) can result in improved optimization in terms of both runtime and log-likelihood.
+
+.. _optimization:
+
+Optimization
+^^^^^^^^^^^^
+
+In addition to being multi-band, the other main difference between *scarlet* and the single-band deblender is that the *scarlet* models are iteratively optimized using a gradient-descent-like algorithm.
+The extra time to run this optimizer is the primary reason that *scarlet* is not used on the single-visit exposures.
+We have tested a number of different algorithms over the years and so far we have achieved the best performance using proxmial ADAM, as described in `Melchior et al. 2019 <https://arxiv.org/abs/1910.10094>`_.
```
taranu: proxmial vs proximal

doc/lsst.meas.extensions.scarlet/overview.rst
```diff
@@ -0,0 +1,217 @@
...
+
+For example, the blend above can be fit in 11 iterations to get the following results, using the *scarlet* models to re-apportion the flux from the observation in each band:
+
+.. image:: images/scarletModel0.png
+.. image:: images/scarletModel1.png
+.. image:: images/scarletModel2.png
+.. image:: images/scarletModel3.png
+
+While much improved, the models still aren't perfect.
+Notice that the *scarlet* model for source 0 has stray flux to the upper left and lower right that (fortunately) doesn't matter in the flux re-apportioned model, however sources 1 and 2 have their flux slightly truncated due the the presence of their neighbor in both the *scarlet* models and flux re-apportioned models.
```
taranu, line 211: How/why does the stray flux end up like that? I would understand if it were in the direction of another bright source but it seems to go perpendicular to neighbours instead.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -442,7 +448,7 @@ def updateBlendRecords(blendData, catalog, modelPsf, observedPsf, redistributeIm
         contains those columns.
     """
     # We import here to avoid a circular dependency
```
taranu, line 453: This seems like code smell...

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -356,6 +358,10 @@ def parse_obj(cls, data: dict) -> "ScarletModelData":
             int(id): ScarletBlendData.fromDict(blend)
             for id, blend in data['blends'].items()
         }
+        if "filters" in dataShallowCopy:
+            # Support the original version,
+            # which used "filters" instead of the now canonical "bands."
+            dataShallowCopy["bands"] = dataShallowCopy.pop("filters")
```
taranu, line 367: I presume the pop is deliberate but you could just leave both keys for backwards compatibility if needed.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -648,6 +654,81 @@ def __init__(self, psfs, model_psf, bbox, dtype):
...
+        This is only required if the user desires to perform a warm restart
+        and wants to mask out the pixels outside of the parent footprint
+        similar to when scarlet was executed in the science pipelines.
+
+    Returns
+    -------
+    blend : `scarlet.lite.LiteBlend`
+        The full scarlet model for the blend.
+    """
+    # Import here to prevent circular import
```
taranu, line 719: As above, surely you can refactor these modules to avoid potential circular imports?

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -316,6 +246,89 @@ def deblend(mExposure, footprint, config, spectrumInit):
...
+    mExposure,
+    footprint=None,
+    badPixelMasks=None,
+    useWeights=True,
+    convolutionType="real",
+):
+    """Generate a LiteObservation from a set of parameters.
+
+    Make the generation and reconstruction of a scarlet model consistent
+    by building either a `DummyObservation` or `LiteObservation` from a
```
taranu: It only seems to return a `LiteObservation`, though?

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -316,6 +246,89 @@ def deblend(mExposure, footprint, config, spectrumInit):
...
+    set of parameters. Because this method serves a dual purpose, some
+    parameters (`mExposure`, `footprint`, `badPixelMasks`) are optional
+    when the observation is only used to reconstruct a model,
+    while others (`modelBox`, `observedPsfs`, `dtype`) are automatically
+    generated from `mExposure`, so they are unnecessary when building a
+    full observation.
```
taranu: If you have two orthogonal sets of parameters that each only apply to one usage, consider making them DataClasses to at least simplify usage and/or docs.

taranu (commented):
A mix of suggestions and mostly minor fixes (typos, etc).

doc/lsst.meas.extensions.scarlet/models.rst
fred3m: Line length. But I re-wrote to use a multi-line list comprehension.

doc/lsst.meas.extensions.scarlet/models.rst
fred3m: You're right, I think that I somehow mixed up the source numbers. This should have been source 2. 

I agree that we shouldn't go through everything in detail, I just wanted to highlight some of the known issues that people will see when looking at deblender results. So I thought that it was important to point out that source 2 (and other more complicated galaxies) will not be modeled well. I'll try to make this section more clear.

doc/lsst.meas.extensions.scarlet/overview.rst
fred3m: I re-worded this but left in the description of what the model is, since I feel like that's relevant information that most people don't know about the SDSS-style deblender.

doc/lsst.meas.extensions.scarlet/overview.rst
fred3m: Oh, I was wrong. It looks like there isn't a config option for this even though the code exists. There probably should be then... DM-33710 will see a much needed refactor of `meas_deblender`, so this will be included in that issue.

doc/lsst.meas.extensions.scarlet/overview.rst
fred3m: I was trying to avoid making the `meas_extensions_scarlet` docs contain too much information about `meas_deblender`, but I guess having a few config parameters listed explicitly isn't the worst thing.

doc/lsst.meas.extensions.scarlet/overview.rst
fred3m: Agreed. This is part of the SDSS algorithm and should only use the PSF _as a template_ for stars, or galaxies that are too faint to have reasonable sizes that can be distinguished from the noise. You can think of it as saying that if a source exists, then it must be _at least_ as large as the PSF, so the PSF template during flux re-distribution.

doc/lsst.meas.extensions.scarlet/overview.rst
fred3m: Nope.

doc/lsst.meas.extensions.scarlet/overview.rst
fred3m: Scarlet lite doesn't have any documentation yet. In the coming months it will be its own stand alone package with additional tests and docs, but for these _are_ the scarlet lite docs.

doc/lsst.meas.extensions.scarlet/overview.rst
fred3m: Example: star forming regions in one part of the galaxies are spectrally similar to star forming regions in other parts of the galaxy since they are made up of the same population of stars.

doc/lsst.meas.extensions.scarlet/overview.rst
fred3m: Updated the typo.
Symmetry is no longer used, so galaxies do not have to be symmetric anymore, but it must be possible to decompose them into a small number of monotonic components. Gaseous regions like nebulae and irregulars break that assumptions because there is no central monotonic region, so even their "center" is ill-defined in terms of where to put the scarlet model.

doc/lsst.meas.extensions.scarlet/overview.rst
fred3m: I'm pretty sure that this would even be true if we were doing direct space convolutions (without over-sampling).

doc/lsst.meas.extensions.scarlet/overview.rst
fred3m: True, but I think this is more than that. The Adaprox algorithm that is being used right now actually prefers solutions further from the true minimum and will jump out of a good solution in the first iteration. This is different than the proximal gradient method (PGM) and other optimization algorithms, which usually step closer in the gradient direction, even in the first step. I've been doing some testing lately and Adaprox seems even more finicky than I had previously thought, and with the PIFF PSFs seems to perform worse than the standard PGM. But that's a discussion for another day...

doc/lsst.meas.extensions.scarlet/overview.rst
fred3m: Your editing ability is much appreciated. I _never_ catch things like this.

doc/lsst.meas.extensions.scarlet/overview.rst
fred3m: From investigations in DM-36324 it looks like the truncate PIFF PSFs are wreaking havoc with our deblender models in a way that I did not appreciate. We saw similar effects in the PSF sources, but for much brighter sources. At first we thought it was an issue with the optimizer but now, with more data, it looks like the problem is that PSF isn't large enough, which means there is no feasible monotonic model to capture all of the flux in brighter sources, so the optmizer gets stuck. I'm working on investigating this now and there will probably be an upcoming talk about it.

python/lsst/meas/extensions/scarlet/io.py
fred3m: This is an artifact of the current state of the code and is not worth fixing. Before the scarlet models were exported, the only place to make measurements on the deconvolved models was in the `ScarletDeblendTask`. But these are band-dependent measurements and the output from `ScarletDeblendTask` is now a single monochromatic catalog. DM-34558 will create a sub-task to make these (and potentially other) measurements, but until that is created there is really no appropriate place for these measurements to live.

python/lsst/meas/extensions/scarlet/io.py
fred3m: In general I agree and most of the refactoring in this ticket was to clean things like this up. However in this case I think that it would add an unnecessary module containing only a single function.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: Yup, that functionality must have been stripped during development. Good catch.

doc/lsst.meas.extensions.scarlet/overview.rst
taranu: Well, yes, but FFT convolution requires zero-padding the image, which I imagine can get quite memory intensive for a large blend without using sparse matrices. Direct convolution doesn't need padding or even a full-sized kernel image, amongst other things.

doc/lsst.meas.extensions.scarlet/overview.rst
taranu: Ok, then perhaps just mention that this is optimizer-dependent and not some quirk of other aspects of Scarlet.

doc/lsst.meas.extensions.scarlet/overview.rst
taranu: Ok, then again I would note that this seems to be the optimizer's fault, rather than a problem with model selection or anything else the user could try to fix.

doc/lsst.meas.extensions.scarlet/overview.rst
taranu: Ok, it sounded ambiguous so I suggest changing it to "the range of colors within each population is not that large", and potentially add something like "e.g. star-forming regions are blue, and bulges tend to be red".

taranu (commented):
A few more minor suggestions for clarification.

PR #59, opened 2022-09-07
DM-35722: Catch errors in setting the HeavyFootprint peak flux

fred3m:


python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -508,7 +510,15 @@ def updateBlendRecords(blendData, catalog, modelPsf, observedPsf, redistributeIm
...
             peak = heavy.peaks[0]
             img = heavyFootprintToImage(heavy, fill=0.0)
-            sourceRecord.set("deblend_peak_instFlux", img.image[Point2I(peak["i_x"], peak["i_y"])])
+            try:
+                sourceRecord.set("deblend_peak_instFlux", img.image[Point2I(peak["i_x"], peak["i_y"])])
+            except Exception:
+                srcId = sourceRecord.getId()
+                x = peak["i_x"]
+                y = peak["i_y"]
+                warnings.warn(f"Source {srcId} at {x},{y} could not set the peak flux with error:")
```
hsinfang: Would it make more sense to use LSST's logging rather than warnings?

hsinfang (approved):

python/lsst/meas/extensions/scarlet/io.py
timj: I agree this looks more like a `log.warning()` -- warnings to get converted to log messages but they won't be found as easily because the wrong logger will be used.

fred3m: Does it work to import the logger using
```
import logging
logger = logging.getLogger("lsst.meas.extensions.scarlet")
logger.warning("Log message here")
```
or do I have to pass it the logger from a task?

timj: I thought measurement plugins had access to a logger so does `logging.getLogger(self.logName)` work to get the logger? That seems to work in some meas_extensions packages.

If that doesn't work use `logger = logging.getLogger(__name__)` so you don't have to hard code the full name in there.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -508,7 +512,15 @@ def updateBlendRecords(blendData, catalog, modelPsf, observedPsf, redistributeIm
...
             img = heavyFootprintToImage(heavy, fill=0.0)
-            sourceRecord.set("deblend_peak_instFlux", img.image[Point2I(peak["i_x"], peak["i_y"])])
+            try:
+                sourceRecord.set("deblend_peak_instFlux", img.image[Point2I(peak["i_x"], peak["i_y"])])
+            except Exception:
+                srcId = sourceRecord.getId()
+                x = peak["i_x"]
+                y = peak["i_y"]
+                logger.warn(f"Source {srcId} at {x},{y} could not set the peak flux with error:")
+                traceback.print_exc()
```
timj, line 522: This should not be called because it won't go to the logging system and we can't tell where it will turn up. Instead you should use `logger.exception()` and that would print your log message above and also include the exception traceback.

PR #70, opened 2023-04-17
DM-38432: Implement scarlet_lite stand alone package

fred3m:


python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -1,476 +1,235 @@
...
-        result : `ScarletSourceData`
-            The reconstructed object
-        """
-        dataShallowCopy = dict(data)
-        del dataShallowCopy["factorized"]
-        components = []
-        for component in data['components']:
-            component = ScarletComponentData.fromDict(component)
+    sources = []
+    # Use a dummy band, since we are only extracting a monochromatic mode
```
taranu: monochromatic model

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -1,476 +1,235 @@
...
+    # that will be turned into a HeavyFootprint.
+    bands = ("band", )
```
taranu: Below you used `dummy` rather than `band`; you may as well be consistent.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -1,476 +1,235 @@
...
-        dataShallowCopy['components'] = components
 
-        factorized = []
-        for component in data["factorized"]:
-            component = ScarletFactorizedComponentData.fromDict(component)
-            factorized.append(component)
-        dataShallowCopy['factorizedComponents'] = factorized
-        dataShallowCopy["peakId"] = int(data["peakId"])
-        return cls(**dataShallowCopy)
+        for factorizdedData in sourceData.factorized_components:
```
taranu: `factorizdedData` -> `factorizedData`

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -1,476 +1,235 @@
...
+            bbox = Box(factorizdedData.shape, origin=factorizdedData.origin)
+            # Add dummy values for properties only needed for
+            # model fitting.
+            spectrum = FixedParameter(factorizdedData.spectrum)
+            totalBands = len(spectrum.x)
+            morph = FixedParameter(factorizdedData.morph)
+            factorized = FactorizedComponent(
+                bands=("dummy", ) * totalBands,
```
taranu, line 89: Is it surprising that the band name not matching "band" above didn't cause any problems, or are band names just labels that get ignored assuming that users keep bands in a consistent order?

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -1,476 +1,235 @@
...
-class ScarletBlendData:
-    """Data for an entire blend.
 
-    Note that `xy0`, `extent`, and `psfCenter` use lsst ``(x, y)`` convention,
-    not the scarlet/C++ ``(y, x)`` convention.
+def updateCatalogFootprints(
+    modelData: scl.io.ScarletModelData,
+    catalog: SourceCatalog,
+    band: str,
+    redistributeImage: afwImage | None = None,
```
taranu: This kind of sounds more like a boolean arg. Maybe `redistributingImage` instead?

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -539,373 +305,64 @@ def updateBlendRecords(blendData, catalog, modelPsf, observedPsf, redistributeIm
...
-
-class DummyObservation(LiteObservation):
-    """An observation that does not have any image data
-
-    In order to reproduce a model in an observed seeing we make use of the
-    scarlet `LiteObservation` class, but since we are not fitting the model
-    to data we can use empty arrays for the image, variance, and weight data,
-    and zero for the `noise_rms`.
+    Note: This is the old version of scarlet, which is deprecated but kept
+    in order to test against th new lsst.scarlet.lite package.
```
taranu: the new

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -539,373 +305,64 @@ def updateBlendRecords(blendData, catalog, modelPsf, observedPsf, redistributeIm
...
-
-
-class DummyObservation(LiteObservation):
-    """An observation that does not have any image data
-
-    In order to reproduce a model in an observed seeing we make use of the
-    scarlet `LiteObservation` class, but since we are not fitting the model
-    to data we can use empty arrays for the image, variance, and weight data,
-    and zero for the `noise_rms`.
+    Note: This is the old version of scarlet, which is deprecated but kept
```
taranu: You mean the old fork-of-scarlet main, right? Not pmelchior/scarlet main itself.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -539,373 +305,64 @@ def updateBlendRecords(blendData, catalog, modelPsf, observedPsf, redistributeIm
...
-        )
-
-        sourceData = ScarletSourceData(
-            components=[componentData],
-            factorizedComponents=[],
-            peakId=source.peakId,
+                components.append(componentData)
+        sourceData = scl.io.ScarletSourceData(
+            components=components,
+            factorized_components=factorizedComponents,
```
taranu, line 419: Are you intending to move more towards snake_case, and if so should I point out possibly old camelCase names or just leave you to convert them incrementally?

python/lsst/meas/extensions/scarlet/metrics.py
```diff
@@ -0,0 +1,105 @@
...
+from lsst.scarlet.lite import Blend
+
+
+@dataclass
+class DeblenderMetrics:
+    """Metrics and measurements made on single sources.
+
+    Store deblender metrics to be added as attributes to a scarlet source
+    before it is converted into a `SourceRecord`.
+    When DM-34414 is finished this class will be eliminated and the metrics
```
taranu: Consider adding TODO.

python/lsst/meas/extensions/scarlet/metrics.py
```diff
@@ -0,0 +1,105 @@
...
+    will be added to the schema using a pipeline task that calculates them
+    from the stored deconvolved models.
+
+    All of the parameters are one dimensional numpy arrays,
+    with an element for each band in the observed images.
+
+    `maxOverlap` is useful as a metric for determining how blended a source
```
taranu, line 46: Isn't it a noisier single-pixel measurement compared to `fluxOverlapFraction`, though?

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -108,140 +101,122 @@ def isPseudoSource(source, pseudoColumns):
...
-        fallback=config.fallback,
-        silent=config.catchFailures,
-        set_spectra=spectrumInit,
+    # Initialize each source with its best fit spectrum
+    if spectrumInit:
+        blend.fit_spectra()
+
+    # Set the optimizer
+    if config.optimizer == "adaprox":
+        # Note: if this doesn't work properly
```
taranu: If this doesn't work properly... then what?

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -108,140 +101,122 @@ def isPseudoSource(source, pseudoColumns):
...
-        iterations = len(blend.loss)
-        failedSources = []
-        for k, src in enumerate(sources):
-            if np.any(~np.isfinite(src.get_model())):
-                failedSources.append(k)
-        raise ScarletGradientError(iterations, failedSources)
+    for k, center in enumerate(peaks):
+        # This is just to make sure that there isn't a coding bug
+        if len(sources[k].components) > 0 and np.any(sources[k].center != center):
+            raise ValueError("Misaligned center, expected {center} but got {sources[k].center}")
```
taranu: Missed the f in front of this f-string

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -527,8 +429,8 @@ class ScarletDeblendConfig(pexConfig.Config):
         dtype=str,
         default="lite",
         allowed={
-            "scarlet": "main scarlet version (likely to be deprecated soon)",
```
taranu, line 576: You should probably think about doing more formal deprecations after this ticket.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -964,8 +866,13 @@ def deblend(self, mExposure, catalog):
                 mask.addMaskPlane(self.config.notDeblendedMask)
 
         # Initialize the persistable data model
-        modelPsf = lite.integrated_circular_gaussian(sigma=self.config.modelPsfSigma)
-        dataModel = ScarletModelData(filters, modelPsf)
+        modelPsf = scl.utils.integrated_circular_gaussian(sigma=self.config.modelPsfSigma)
+        dataModel = scl.io.ScarletModelData(filters, modelPsf)
+
+        # Initialize the monotonicity operator with a size of 101 x 101 pixels.
+        # Note: If a component is > 101x101 then the monotonicity operator
```
taranu: Greater in either or both axes, and resize in which direction and for which axes?

python/lsst/meas/extensions/scarlet/utils.py
```diff
@@ -31,3 +47,201 @@ def footprintsToNumpy(catalog, shape, xy0=None):
         yidx, xidx = spans.shiftedBy(*offset).indices()
         result[yidx, xidx] = 1
     return result
+
+
+def scarletBoxToBBox(box: scl.Box, xy0: geom.Point2I = geom.Point2I()) -> geom.Box2I:
```
taranu, line 52: So after this ticket, any references to "scarlet" will mean "scarlet_lite" and not pmelchior/scarlet main unless otherwise specified, right?

python/lsst/meas/extensions/scarlet/utils.py
```diff
@@ -31,3 +47,201 @@ def footprintsToNumpy(catalog, shape, xy0=None):
...
+    """Convert a scarlet_lite Box into a Box2I.
+
+    Parameters
+    ----------
+    box:
+        The scarlet bounding box to convert.
+    xy0:
+        An additional offset to apply to the scarlet box.
```
taranu: Suggest the unambiguous "to add to" rather than "to apply to" (also to mirror the doc in `bboxToScarletBox`).

python/lsst/meas/extensions/scarlet/utils.py
```diff
@@ -31,3 +47,201 @@ def footprintsToNumpy(catalog, shape, xy0=None):
...
+
+def bboxToScarletBox(bbox: geom.Box2I, xy0: geom.Point2I = geom.Point2I()) -> scl.Box:
+    """Convert a Box2I into a scarlet_lite Box.
+
+    Parameters
+    ----------
+    bbox:
+        The Box2I to convert into a scarlet `Box`.
+    xy0:
+        An overall offset to subtract from the `Box2I`.
```
taranu, line 84: This may be more confusing than keeping the sign consistent between both methods, but as long as it's documented I suppose that's fine.

python/lsst/meas/extensions/scarlet/utils.py
```diff
@@ -31,3 +47,201 @@ def footprintsToNumpy(catalog, shape, xy0=None):
...
+        `(0,0)` at the lower left corner of the blend while
+        the blend itself is likely to have an offset in the
+        `Exposure`.
+
+    Returns
+    -------
+    bbox:
+        The converted bounding box.
+    """
+    xy0 = geom.Point2I(box.origin[-1]+xy0.x, box.origin[-2]+xy0.y)
```
taranu: Spaces around +

python/lsst/meas/extensions/scarlet/utils.py
```diff
@@ -31,3 +47,201 @@ def footprintsToNumpy(catalog, shape, xy0=None):
...
+        location of the blend and `bbox` is the box containing
+        a source in the blend.
+
+    Returns
+    -------
+    box:
+        A scarlet `Box` that is more useful for slicing image data
+        as a numpy array.
+    """
+    origin = (bbox.getMinY()-xy0.y, bbox.getMinX()-xy0.x)
```
taranu: Spaces around -

python/lsst/meas/extensions/scarlet/utils.py
```diff
@@ -31,3 +47,201 @@ def footprintsToNumpy(catalog, shape, xy0=None):
...
+def buildObservation(
+    modelPsf: np.ndarray,
+    psfCenter: tuple[int, int] | geom.Point2I | geom.Point2D,
+    mExposure: MultibandExposure,
+    badPixelMasks: list[str] | None = None,
+    footprint: afwFootprint = None,
+    useWeights: bool = True,
+    convolutionType: str = "real",
+) -> scl.Observation:
+    """Generate an Observation from a set of parameters.
```
taranu: I'm confused, what are the parameters here? Or do you not mean Scarlet parameters but just the args?

python/lsst/meas/extensions/scarlet/utils.py
```diff
@@ -31,3 +47,201 @@ def footprintsToNumpy(catalog, shape, xy0=None):
...
+        attached to the observation.
+    footprint:
+        The footprint that is being fit.
+        If `footprint` is `None` then the weights are not updated to mask
+        out pixels not contained in the footprint.
+    badPixelMasks:
+        The keys from the bit mask plane used to mask out pixels
+        during the fit.
+        If `badPixelMasks` is `None` then the default values from
+        `ScarletDeblendConfig.badMask` is used.
```
taranu: are used

python/lsst/meas/extensions/scarlet/utils.py
```diff
@@ -31,3 +47,201 @@ def footprintsToNumpy(catalog, shape, xy0=None):
...
+    useWeights:
+        Whether or not fitting should use inverse variance weights to
+        calculate the log-likelihood.
+    convolutionType:
+        The type of convolution to use (either "real" or "fft").
+        When reconstructing an image it is advised to use "real" to avoid
+        polluting the footprint with
```
taranu: ... with negative-valued pixels? Artifacts?

python/lsst/meas/extensions/scarlet/utils.py
```diff
@@ -31,3 +47,201 @@ def footprintsToNumpy(catalog, shape, xy0=None):
...
+    # Initialize the observed PSFs
+    if not isinstance(psfCenter, geom.Point2D):
+        psfCenter = geom.Point2D(*psfCenter)
+    psfModels = mExposure.computePsfKernelImage(psfCenter)
+
+    # Use the inverse variance as the weights
+    if useWeights:
+        weights = 1/mExposure.variance.array
+    else:
+        # Mask out bad pixels
```
taranu: Shouldn't this comment be one line lower? The next line doesn't do any masking.

python/lsst/meas/extensions/scarlet/utils.py
```diff
@@ -31,3 +47,201 @@ def footprintsToNumpy(catalog, shape, xy0=None):
...
+        the observation, PSF, etc, used to convolve the
+        scarlet model to the observed seeing in each band.
+    useFlux:
+        Whether or not to re-distribute the flux from the image
+        to conserve flux.
+
+    Returns
+    -------
+    heavy:
+        The footprint (possibly multiband) containing the model for the source.
```
taranu, line 200: Is `MultibandFootprint` actually a `HeavyFootprintF`? If not, the return value type hint should be updated.

tests/test_deblend.py
```diff
@@ -168,28 +179,31 @@ def test_deblend_task(self):
...
-                    position = Point2D(*blendData.psfCenter)
-                    _psfs = coadds[band].getPsf().computeKernelImage(position).array[None, :, :]
-                    modelBox = Box((1,) + tuple(blendData.extent[::-1]), origin=(0, 0, 0))
+                    position = Point2D(*blendData.psf_center[::-1])
+                    _psfs = self.coadds[band].getPsf().computeKernelImage(position).array[None, :, :]
+                    # The stored PSF should be the same as the calculated one
+                    assert_almost_equal(blendData.psf[bandIndex:bandIndex+1], _psfs)
+                    modelBox = scl.Box(blendData.shape, origin=blendData.origin)
                     blend.observation = DummyObservation(
+                        bands=("band", ),
```
taranu: Consider defining a consistent dummy band name somewhere in the package (for convenience, if nothing else).

tests/test_deblend.py
```diff
@@ -234,6 +256,61 @@ def test_deblend_task(self):
         skipped = largeFootprint | denseFootprint
         np.testing.assert_array_equal(skipped, catalog["deblend_skipped"])
 
+    def test_continuity(self):
+        """This test ensures that lsst.scarlet.lite gives roughly the same
+        result as scarlet.lite
+
+        This test can be removed once the deprecated scarlet.lite
```
taranu: Consider adding a TODO (definitely if the ticket already exists).

tests/test_deblend.py
```diff
@@ -234,6 +256,61 @@ def test_deblend_task(self):
...
+        """
+        oldCatalog, oldModelData, oldConfig = self._deblend("old_lite")
+        catalog, modelData, config = self._deblend("lite")
+
+        # Ensure that the deblender used different versions
+        self.assertEqual(oldConfig.version, "old_lite")
+        self.assertEqual(config.version, "lite")
+
+        # Check that the PSF and other properties are the same
+        assert_almost_equal(oldModelData.psf, modelData.psf)
```
taranu, line 272: Are you expecting machine epsilon level differences or larger ones (in which case this might be too sensitive of a check)?

tests/test_deblend.py
```diff
@@ -234,6 +256,61 @@ def test_deblend_task(self):
...
+        for blendId in modelData.blends.keys():
+            oldBlendData = oldModelData.blends[blendId]
+            blendData = modelData.blends[blendId]
+
+            # Check that blend properties are the same
+            self.assertTupleEqual(oldBlendData.origin, blendData.origin)
+            self.assertTupleEqual(oldBlendData.shape, blendData.shape)
+            self.assertTupleEqual(oldBlendData.psf_center, blendData.psf_center)
+            self.assertTupleEqual(tuple(oldBlendData.sources.keys()), tuple(blendData.sources.keys()))
+            assert_almost_equal(oldBlendData.psf, blendData.psf)
```
taranu, line 289: This is the PSF per blend, which is different from the `oldModelData.psf` above, right?

taranu (commented):
Broadly fine, but I have some more clarification requests and small fixes.

python/lsst/meas/extensions/scarlet/io.py
fred3m: The labels are meaningless when the model is only used in a single band. We only care about the band names if we are looking at the full multiband models.

python/lsst/meas/extensions/scarlet/io.py
fred3m: Changed to `imageForRedistribution`.

python/lsst/meas/extensions/scarlet/io.py
fred3m: I did mean the old version, but I made it clear that this converts data from the old scarlet into the persistable data format in the new scarlet lite package.

python/lsst/meas/extensions/scarlet/io.py
fred3m: It's actually the same for `scarlet.lite.` But the new comment clarifies.

python/lsst/meas/extensions/scarlet/io.py
fred3m: scarlet_lite uses snake case, but I'm going to keep `meas_extensions_scarlet` in camel case both to minimize work and keep it consistent with other measurement packages.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: 😆 I think this was an old comment use when I was developing this and I only deleted part of the comment.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: Agreed. In this case `version=scarlet` was only for testing and was never used anywhere in production. I would also argue that the `oldScarletToData` can be safely removed once this PR is accepted since it only exists to validate this ticket.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: I clarified in the comment, but point of the new operator is that it starts out with small dimensions for the largest source and resizes itself whenever a larger source exists.

python/lsst/meas/extensions/scarlet/utils.py
fred3m: This is mostly irrelevant now. In "old" scarlet, and scarlet lite, the bounding box was always from (0,0)-(height, width), so the offset was always added to the origin of the `scarlet.Box`. In scarlet_lite the bbox has the same origin as the box2I from the footprint, so this is typically zero.

Do you think that I should just remove it now?

python/lsst/meas/extensions/scarlet/utils.py
fred3m: Yes, clarified to state that it uses the args.

python/lsst/meas/extensions/scarlet/utils.py
fred3m: Updated by adding "... with artifacts from the fft."

python/lsst/meas/extensions/scarlet/utils.py
fred3m: It can be either one, thanks for pointing this out. I updated the docstring and type hint.

tests/test_deblend.py
fred3m: Yes, switched to "dummy" for the entire package.

tests/test_deblend.py
fred3m: These should be identical, but I made them machine level just to be safe.

tests/test_deblend.py
fred3m: Yes. `oldModelData.psf` and `modelData.psf` is the model PSF that all of the blends are generated in.

PR #72, opened 2023-09-13
DM-40451: Allow deblending to continue even when data is missing for some bands

fred3m:
Previously any footprints that could not construct a PSF in all bands were skipped. As part of HSC PDR4, because there are patches with only partial coverage of narrow bands (for example), we want to allow the deblender to proceed.

This ticket implements deblending with only partial PSF coverage, and also adds new columns to the catalog to keep track of the coverage per band.

### DIFF
--- a/python/lsst/meas/extensions/scarlet/io.py
+++ b/python/lsst/meas/extensions/scarlet/io.py
@@ -367,7 +367,7 @@ class ScarletModelData:
         }
         return cls(**dataShallowCopy)
 
-    def updateCatalogFootprints(self, catalog, band, psfModel, maskImage, redistributeImage=None,
+    def updateCatalogFootprints(self, catalog, band, psfModel, maskImage=None, redistributeImage=None,
                                 removeScarletData=True, updateFluxColumns=True):
         """Use the scarlet models to set HeavyFootprints for modeled sources
 
@@ -382,6 +382,8 @@ class ScarletModelData:
         maskImage : `lsst.afw.image.MaskX`
             The masked image used to calculate the fraction of pixels
             in each footprint with valid data.
+            This is only used when `updateFluxColumns` is `True`,
+            and is required if it is.
         redistributeImage : `lsst.afw.image.Image`
             The image that is the source for flux re-distribution.
             If `redistributeImage` is `None` then flux re-distribution is
@@ -489,6 +491,8 @@ def updateBlendRecords(blendData, catalog, modelPsf, observedPsf, maskImage, red
     maskImage : `lsst.afw.image.MaskX`
         The masked image used to calculate the fraction of pixels
         in each footprint with valid data.
+        This is only used when `updateFluxColumns` is `True`,
+        and is required if it is.
     redistributeImage : `lsst.afw.image.Image`
         The image that is the source for flux re-distribution.
         If `redistributeImage` is `None` then flux re-distribution is
@@ -558,11 +562,12 @@ def updateBlendRecords(blendData, catalog, modelPsf, observedPsf, maskImage, red
             useFlux=useFlux,
         )
         sourceRecord.setFootprint(heavy)
-        # Set the fraction of pixels with valid data.
-        coverage = calculateFootprintCoverage(heavy, maskImage)
-        sourceRecord.set("deblend_dataCoverage", coverage)
 
         if updateFluxColumns:
+            # Set the fraction of pixels with valid data.
+            coverage = calculateFootprintCoverage(heavy, maskImage)
+            sourceRecord.set("deblend_dataCoverage", coverage)
+
             # Set the flux of the scarlet model
             # TODO: this field should probably be deprecated,
             # since DM-33710 gives users access to the scarlet models.
--- a/python/lsst/meas/extensions/scarlet/io.py
+++ b/python/lsst/meas/extensions/scarlet/io.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 import json
-from typing import Any
 import logging
 import numpy as np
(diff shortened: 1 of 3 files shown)


python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -422,7 +444,35 @@ def updateCatalogFootprints(self, catalog, band, psfModel, redistributeImage=Non
...
+    maskImage : `lsst.afw.image.MaskX`
+        The mask image with the ``NO_DATA`` bit set.
+
+    Returns
+    -------
+    coverage : `float`
+        The fraction of pixels in `footprint` where the ``NO_DATA`` bit is set.
+    """
+    # Store the value of "NO_DATA" from the mask plane.
+    noDataInt = 2**maskImage.getMaskPlaneDict()["NO_DATA"]
```
cmsaunders, line 465: You could just do `noDataInt = maskImage.getPlaneBitMask("NO_DATA")` here.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -238,7 +268,7 @@ def deblend(mExposure, footprint, config, spectrumInit):
     # Store the location of the PSF center for storage
     blend.psfCenter = (psfCenter.x, psfCenter.y)
 
-    return blend, skipped
+    return blend, skipped, []
```
cmsaunders: Can you add a comment explaining this? Also, I saw that the docstrings say `spectrumInit` is returned, but it looks like it's not. 
Edit: Okay, I see that you have this to match the output of `deblend_lite`, but a comment plus corrected docstrings would still be helpful.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -108,6 +108,36 @@ def isPseudoSource(source, pseudoColumns):
...
+    mExposure : `MultibandExposure`
+        The exposure, updated to only use bands that
+        successfully generated a PSF image.
+    """
+    if not isinstance(psfCenter, geom.Point2D):
+        psfCenter = geom.Point2D(*psfCenter)
+
+    try:
+        psfModels = mExposure.computePsfKernelImage(psfCenter)
+    except IncompleteDataError as e:
```
cmsaunders, line 133: My general feeling is that try-excepts should be avoided as much as possible. Since you are already changing the parameters and returned values in `computePsfKernelImage`, you could consider adding something like an `allowMissing` option, then always returning the psfs and the filters used. I'm not sure if that would be simpler.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -819,6 +865,10 @@ def _addSchemaKeys(self, schema):
                                                 doc="Deblending failed on source")
         self.deblendErrorKey = schema.addField('deblend_error', type="String", size=25,
                                                doc='Name of error if the blend failed')
+        self.incompleteDataKey = schema.addField('deblend_incompleteData', type='Flag',
+                                                 doc='True when a bland has at least one band '
```
cmsaunders: `bland` -> `blend`

tests/test_deblend.py
```diff
@@ -104,6 +104,7 @@ def test_deblend_task(self):
         src.setFootprint(denseFoot)
 
         # Run the deblender
+        print("RUNNING!")
```
cmsaunders: I assume this was for debugging?

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -1009,9 +1061,9 @@ def deblend(self, mExposure, catalog):
                 t0 = time.monotonic()
                 # Build the parameter lists with the same ordering
                 if self.config.version == "scarlet":
-                    blend, skipped = deblend(mExposure, foot, self.config, spectrumInit)
+                    blend, skippedSources, skippedBands = deblend(mExposure, foot, self.config, spectrumInit)
```
cmsaunders: Continuing the thought from above, you could have `deblend` just return `blend` and `skippedSources`, then set `skippedBands = []` here, so that the `deblend` outputs aren't forced to match `deblend_lite`.

cmsaunders (approved):
This looks good. I have a few comments about places where I think you could make the logic more straightforward, but it's your call whether you make those changes.

python/lsst/meas/extensions/scarlet/io.py
fred3m: Oooo, nice!

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: I think this is a good example of a case where it is easier to ask forgiveness than acceptance. I originally used an `allowMissing` parameter but it requires an additional check by the user to see if any bands were missing. Since this check is already being done while creating the PSF image I wanted to find a way to avoid having to validate the output again, and this seemed like a good way to ensure that users who don't expect a bad PSF will get an error, and users who know that it might happen do not have to check for it.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: Yeah, that does make more sense.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
cmsaunders: Ok, sounds reasonable to me.

PR #76, opened 2023-10-02
DM-41008: Update deblend_coverage for all parents

fred3m:


python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -153,13 +153,18 @@ def updateCatalogFootprints(
             # no models for its sources.
             continue
 
+        parent = catalog.find(parentId)
+        if updateFluxColumns and imageForResdistibution is not None:
```
ktlim: Is it worth fixing this to `imageForRedistribution` now?

fred3m: Sure, although I'll have to update a few more packages now...

hsinfang (approved):

PR #93, opened 2024-03-08
DM-43176: Flag sources with no flux after re-distribution.

fred3m:
Sometimes errors in the pipelines, like incorrect PSF modeling, can cause sources with valid scarlet models to have no flux after flux re-distribution when the footprints are being reconstructed. This commit adds a `deblend_zeroFlux` column to identify sources that have zero flux in a given band for follow-up.

### DIFF
--- a/python/lsst/meas/extensions/scarlet/io.py
+++ b/python/lsst/meas/extensions/scarlet/io.py
@@ -26,7 +26,8 @@ import numpy as np
 
 from lsst.afw.table import SourceCatalog
 from lsst.afw.image import MaskedImage, Exposure
-from lsst.afw.detection import Footprint as afwFootprint
+from lsst.afw.detection import Footprint as afwFootprint, HeavyFootprintF
+from lsst.afw.geom import SpanSet, Span
 from lsst.geom import Box2I, Extent2I, Point2I
 import lsst.scarlet.lite as scl
 from lsst.scarlet.lite import Blend, Source, Box, Component, FixedParameter, FactorizedComponent, Image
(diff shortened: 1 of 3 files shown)


python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -333,9 +334,25 @@ def updateBlendRecords(
...
+                sourceRecord.set("deblend_zeroFlux", True)
+                # Create a Footprint with a single pixel, set to zero,
+                # to avoid breakage in measurement algorithms.
+                dtype = observation.images.dtype
+                center = Point2I(heavy.peaks[0]["i_x"], heavy.peaks[0]["i_y"])
+                spanList = [Span(center.y, center.x, center.x)]
+                footprint = afwFootprint(SpanSet(spanList))
+                footprint.setPeakCatalog(heavy.peaks)
+                zeroImage = afwImage(array=np.array([[0]], dtype=np.float32), xy0=center, dtype=dtype)
+                heavy = makeHeavyFootprint(footprint, MaskedImage(zeroImage, dtype=dtype))
```
TallJimbo:
```suggestion
                heavy = HeavyFootprintF(footprint)
                heavy.getImageArray()[0] = 0.0
```
is probably a bit more efficient (no need for `Image` or `MaskedImage` temporaries), but I'm not sure it's enough better to merit another Jenkins run.  Just another way to do it.

TallJimbo (approved):

python/lsst/meas/extensions/scarlet/io.py
fred3m: Good idea. I have to rebase the other packages anyway, so I'll make this change, rebase all of them, and just Jenkins without the ci packages.

PR #94, opened 2024-03-13
Implement DM-43176 to skip zero flux sources

fred3m:


fred3m:
It was decided not to implement backports for this ticket because they change the schema by adding a new column.

PR #100, opened 2024-09-20
DM-41840: Implement new detection tasks

fred3m:


python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
```diff
@@ -0,0 +1,301 @@
...
+    )
+
+
+class DeconvolveExposureConfig(
+    pipeBase.PipelineTaskConfig,
+    pipelineConnections=DeconvolveExposureConnections,
+):
+    """Configuration for DeconvolveExposureTask"""
+
+    maxIter = pexConfig.Field(
```
taranu: `pexConfig.Field[int]` (here and elsewhere)

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
```diff
@@ -0,0 +1,301 @@
...
+    coadd = cT.Input(
+        doc="Exposure to deconvolve",
+        name="{inputCoaddName}Coadd_calexp",
+        storageClass="ExposureF",
+        dimensions=("tract", "patch", "band", "skymap"),
+    )
+
+    peaks = cT.Input(
+        doc="Catalog of detected peak positions",
+        name="{inputCoaddName}_coadd_multiband_peaks",
```
taranu: Is this in anticipation of [RFC-1049](https://rubinobs.atlassian.net/browse/RFC-1049)? I debated this myself; it might be premature to move to full snake case before it's implemented and `deepCoadd` changes.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
```diff
@@ -0,0 +1,301 @@
...
+    )
+    useWavelets = pexConfig.Field(
+        dtype=bool,
+        doc="Deconvolve using wavelets to supress high frequency noise",
+        default=True,
+    )
+    waveletGeneration = pexConfig.Field(
+        dtype=int,
+        default=2,
+        doc="Generation of the starlet wavelet used for peak detection (should be 1 or 2). "
```
taranu: Why not make it a ChoiceField then? Or at least validate the value.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
```diff
@@ -0,0 +1,301 @@
...
+        "Only used if useWavelets is True",
+    )
+    waveletScales = pexConfig.Field(
+        dtype=int,
+        default=1,
+        doc="Number of wavelet scales used for peak detection. Only used if useWavelets is True",
```
taranu: I wish there were a better pattern for this. Letting ConfigFields be null/None would've been nice...

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
```diff
@@ -0,0 +1,301 @@
...
+        # Remove the high frequency wavelets.
+        # This has the effect of preventing high frequency noise
+        # from interfering with the detection of peak positions.
+        wavelets[0] = 0
+        # Reconstruct the image from the remaining wavelet coefficients
+        image = scl.wavelet.starlet_reconstruction(
+            wavelets[:, 0],
+            generation=self.config.waveletGeneration,
+        )
+        # Remove the high frequency wavelets from the PSF.
```
taranu: Does this leave the PSF consistent with the (partially?) deconvolved image? Does that matter?

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
```diff
@@ -0,0 +1,301 @@
...
+        bands = (band,)
+        model_psf = scl.utils.integrated_circular_gaussian(sigma=0.8)
+
+        if self.config.useWavelets:
+            image, psf = self._removeHighFrequencySignal(coadd)
+        else:
+            image = coadd.image.array
+            psf = coadd.getPsf().computeKernelImage(coadd.getBBox().getCenter()).array
+        weights = np.ones_like(coadd.image.array)
+        badPixelMasks = ["SAT", "INTRP", "NO_DATA"]
```
taranu: Hmm, does INTRP actually help here?

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
```diff
@@ -0,0 +1,301 @@
...
+            component_peaks = None
+        else:
+            component_peaks = [(peak["i_y"], peak["i_x"]) for peak in peaks]
+
+        # Initialize the model as a single source with a single component:
+        # the entire image.
+        component = scl.models.free_form.FreeFormComponent(
+            bands=observation.bands,
+            model=observation.images.data.copy(),
+            model_bbox=observation.bbox,
```
taranu: Since you're not extrapolating flux from outside of the bbox (right?), does this this mean that you're likely to overestimate the denconvolved flux within half of the PSF dimensions from the edge of the bbox?

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
```diff
@@ -0,0 +1,268 @@
...
+    )
+
+
+class MultiBandDetectionConfig(
+    pipeBase.PipelineTaskConfig,
+    pipelineConnections=MultiBandDetectionConnections,
+):
+    """Configuration for MultiBandDetectionTask"""
+
+    doDetectPeaks = pexConfig.Field(
```
taranu: `Field[bool]` etc.

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
```diff
@@ -0,0 +1,268 @@
...
+        dtype=bool,
+        doc="Detect peaks in the input images. "
+        "If False then the multi-band peak catalog is used.",
```
taranu: Shouldn't you delete the peaks connection if it's `True`, then?

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
```diff
@@ -0,0 +1,268 @@
...
+        default=False,
+    )
+    minFootprintArea = pexConfig.Field(
+        dtype=int,
+        default=4,
+        doc="Minimum area of a footprint to be considered detectable",
```
taranu: Does this need a `>0` check? Same for most of these minima, actually.

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
```diff
@@ -0,0 +1,268 @@
...
+        outputs = self.run(**inputs)
+        butlerQC.put(outputs, outputRefs)
+
+    def run(
+        self,
+        coadds: list[afwImage.Exposure],
+        bands: list[str],
+        idFactory: IdFactory,
+        peaks: PeakCatalog | None = None,
+    ) -> pipeBase.Struct:
```
taranu: Needs a docstring.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
```diff
@@ -0,0 +1,301 @@
...
+        outputs = self.run(**inputs)
+        butlerQC.put(outputs, outputRefs)
+
+    def run(
+        self,
+        coadd: afwImage.Exposure,
+        band: str,
+        peaks: PeakCatalog | None = None,
+        **kwargs
+    ) -> pipeBase.Struct:
```
taranu: Needs a docstring.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
```diff
@@ -0,0 +1,301 @@
...
+        )
+        psf_wavelets[0] = 0
+        psf = scl.wavelet.starlet_reconstruction(
+            psf_wavelets[:, 0],
+            generation=self.config.waveletGeneration,
+        )
+        return image, psf
+
+    def _buildObservation(self, coadd: afwImage.Exposure, band: str):
+        """Build a scarlet lite Observation from an Exposure."""
```
taranu: Probably should have arg docs even if they're self-explanatory.

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
```diff
@@ -0,0 +1,268 @@
...
+        self._attachExternalPeaks(peaks, footprints, sources, mCoadd.getBBox())
+
+        return pipeBase.Struct(footprints=footprints, sources=sources)
+
+    def _detectPeaks(
+        self,
+        mCoadd: afwImage.MultibandExposure,
+        footprints: Sequence[scl.detect.Footprint],
+    ):
+        """Detect peaks in the detection image"""
```
taranu: arg docs

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
```diff
@@ -0,0 +1,268 @@
...
+        # Build the noise weighted detection image.
+        sigma = np.median(np.sqrt(mCoadd.variance.array), axis=(1, 2)) / 2
```
taranu: Why half of the median? Should this be configurable?

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
```diff
@@ -0,0 +1,268 @@
...
+            min_separation=self.config.minPeakDistance,
+            min_area=self.config.minFootprintArea,
+            peak_thresh=self.config.minPeakSNR,
+            footprint_thresh=self.config.minFootprintSNR,
+            find_peaks=True,
+            remove_high_freq=False,
+            min_pixel_detect=1,
+        )
+        peaks = utils.scarletFootprintsToPeakCatalog(wide_footprints)
+        log.debug("Total number of peaks detected:", len(peaks))
```
taranu: I think this would be fine as `info`-level.

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
```diff
@@ -0,0 +1,268 @@
...
+        return peaks
+
+    def _attachExternalPeaks(
+        self,
+        peaks: PeakCatalog,
+        footprints: Sequence[scl.detect.Footprint],
+        sources: SourceCatalog,
+        bbox: Box2I,
+    ):
+        """Attach peaks to the footprints in the source catalog"""
```
taranu: arg docs

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
```diff
@@ -0,0 +1,268 @@
...
+        scarletBox = utils.bboxToScarletBox(bbox)
+        xmin, ymin = bbox.getMin()
+        footprint_image = scl.detect.footprints_to_image(footprints, scarletBox).data
```
taranu: This is a boolean mask of pixels within footprints?

python/lsst/meas/extensions/scarlet/multiBandPeakDetectionTask.py
```diff
@@ -0,0 +1,147 @@
...
+log = logging.getLogger(__name__)
+
+__all__ = [
+    "MultiBandPeakDetectionTask",
+    "MultiBandPeakDetectionConfig",
+    "MultiBandPeakDetectionConnections",
+]
+
+
+class MultiBandPeakDetectionConnections(
```
taranu: Is it really not possible to consolidate this task with `MultiBandDetection`? Maybe the task and connections are trivial but some of the config fields look identical.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -228,21 +236,36 @@ class ScarletDeblendConfig(pexConfig.Config):
...
-    minIter = pexConfig.Field(dtype=int, default=15,
-                              doc="Minimum number of iterations before the optimizer is allowed to stop.")
-    maxIter = pexConfig.Field(dtype=int, default=300,
-                              doc=("Maximum number of iterations to deblend a single parent"))
-    relativeError = pexConfig.Field(dtype=float, default=1e-2,
-                                    doc=("Change in the loss function between iterations to exit fitter. "
-                                         "Typically this is `1e-2` if measurements will be made on the "
-                                         "flux re-distributed models and `1e-4` when making measurements "
-                                         "on the models themselves."))
+    minIter = pexConfig.Field(
```
taranu: `Field[int]` etc.

taranu (commented):
See individual comments.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
fred3m: I think it's fine to do this now. Jim seemed to give his support to this naming scheme: https://lsstc.slack.com/archives/CAVQQ6SBX/p1726691030465659?thread_ts=1726688938.272269&cid=CAVQQ6SBX Besides, I don't see the need to create one more thing that needs to be changed shortly anyway.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
fred3m: 100% agree, but there doesn't seem to be a way around it (and this pattern is used often).

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
fred3m: So that's the insight that I had while writing this algorithm. The first order starlets are the difference between convolving the image with the starlet kernel, so removing the high frequency signal just leaves the starlet convolved with the image. I did a quick writeup yesterday and today to describe the algorithm and how it works and Ive attached it to the ticket.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
fred3m: I'm not sure. These are the same flags that I ignore in ScarletDeblendTask. I think my worry at the time was that poorly interpolated values can make things worse, and because the algorithm does a convolution the neighboring pixels will help constrain any masked pixels, so I'd rather ignore every pixel whose value is called into question.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
fred3m: Yes, that's correct. But this is the bounding box of the entire image (none of these algorithms are per blend), since a "component" is really just a deconvolved model of the entire image in this case. So this will be true of any convolution algorithm.

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
fred3m: Yup, fixed all of these.

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
fred3m: It does.

```
def __init__(self, *, config=None):
        if config.doDetectPeaks:
            # Deconvolution does not use input catalog
            self.inputs.remove("peaks")
```

fred3m: Maybe you started to review this several weeks ago before I caught this error?

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
fred3m: I don't think we typically go so far as to check that counts and areas are positive do we? That seems like overkill for a user error that should be easy to catch.

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
fred3m: No, it's actually an image with an integer (the index of the footprint in the catalog) for each pixel. So `footprint_image[y, x]` gives the index of the footprint at that location. This is actually an idea that Nate and I came up with one day when trying to figure out the best way to insert peaks into existing footprints without having to search the catalogs.

python/lsst/meas/extensions/scarlet/multiBandPeakDetectionTask.py
fred3m: I could, but they're different tasks that do slightly different things, run at different times in the same pipeline. If anything I could make `MultiBandPeakDetectionTask` a subtask of `MultiBandDetectionTask`, so that it can be run on its own before deconvolution and again after deconvolution (if necessary).

If you don't mind I'd like to keep them separate for now, since
a) I don't know if `MultiBandPeakDetectionTask` works as well as I'd hoped (right not it looks like it's having some trouble)
b) It gives me more freedom to tinker with both of them independently while I test things out.

This doesn't mean that they'll always be separate, but for now I think that it's convenient to keep them the way that they are.

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
taranu: There's no `__init__` in `MultiBandDetectionConnections` right now...

fred3m: Hmmm, looks like I did forget to push a commit. But it's updated now.

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
fred3m: I think I was testing something and just forgot to remove that 2. This is a good catch, because it wouldn't make any difference anyway.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
taranu: Alright.. I don't have an intuition for wavelets or starlets so I only half-understood, but other people on the project do so fair enough.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
taranu: Okay then, this is probably worth investigating for both this task and Scarlet deblending once we switch to cell coadds, if you don't already have a ticket for that.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
taranu: I see. The border region won't extend to the inner region of the patch unless the PSF is larger than the overlap then, right?

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
taranu: I suppose it depends on whether you want users to be allowed to set it to zero. Would it actually work as expected then?

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
taranu: Okay, that sounds like a segmentation map then.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
fred3m: Created DM-47571

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
fred3m: Correct.

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
fred3m: I completely removed this task now (explanation coming in the ticket).

python/lsst/meas/extensions/scarlet/multiBandDetectionTask.py
fred3m: Yes, for a number of reasons. The main one is that the algorithm is incapable of create a zero size footprint (because the pixel being checked will always have itself in the footprint).

PR #102, opened 2024-12-11
DM-48097: Update maxFootprintArea from 1e5 to 1e6

fred3m:


kherner (approved):

kherner:
Fred presented the proposed change in the Science Pipelines meeting on 11 December and it was received without objection.

PR #106, opened 2025-03-18
DM-36099: Change 'filters' to 'bands' for multiband images and exposures

fred3m:


python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -518,8 +518,8 @@ def __init__(self, schema, peakSchema=None, **kwargs):
             Any fields beyond the PeakTable minimal schema will be transferred
             to the main source Schema.  If None, no fields will be transferred
             from the Peaks.
-        filters : list of str
-            Names of the filters used for the eposures. This is needed to store
+        bands : list of str
```
arunkannawadi:
```suggestion
        bands : `list` [`str`]
```

since you're changing this line anyway.
https://developer.lsst.io/python/numpydoc.html#sequence-types

arunkannawadi: But anyway, this argument is not in the function signature nor do I see it being used anywhere in the code. May be its time to remove it altogether?

arunkannawadi (approved):

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: Yup, I just removed it. Good catch!

PR #108, opened 2025-04-01
DM-49537: Implement new scarlet model storage API

fred3m:


python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -472,3 +482,153 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+            f'blend_{blend_id}.json': json.dumps(blend_data)
+            for blend_id, blend_data in json_model['blends'].items()
+        }
+        data.update({
+            'psf_shape.json': json.dumps(json_model['psfShape']),
+            'psf.json': json.dumps(json_model['psf']),
+        })
+        return data
+
+    def read_from_uri(self, uri: ResourcePath, component: str | None = None, expected_size: int = -1) -> Any:
```
timj: Consider using `read_from_stream` instead of `read_from_uri` since you are immediately opening a stream anyhow. It would save you opening the file.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -472,3 +482,153 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+        # Override of `FormatterV2.read_from_uri`.
+        if self.file_descriptor.parameters is not None and "blend_id" in self.file_descriptor.parameters:
+            blend_id = self.file_descriptor.parameters["blend_id"]
```
timj: It seems like the infrastructure here could just as easily support a request for multiple blends in parameters rather than a single blend. Is that at all useful? `blend_ids` could be scalar or list of IDs (and `lsst.utils.iteration.ensure_iterable` can unify them to a single list).

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -472,3 +482,153 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+            Dictionary mapping filenames to the desired file type.
+        """
+        results = {}
+        filenames = [
+            blend_filename,
+            'psf.json',
+            'psf_shape.json',
+        ]
+        with uri.open("rb") as f:
+            with zipfile.ZipFile(f, 'r') as zipf:
```
timj: It seems like _extract_blend and _extract_files could be unified as something like:
```python
default_files = []
if blend_filename:
    default_files = [blend_filename, "psf.json", "psf_shape.json"]
results = {}
with zipfile.ZipFile(f, "r") as zipf:
    filenames = default_files or zipf.namelist()
    for filename in filenames:
        ...
return results
```
Then if blend_filename is None it reads all of them.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -472,3 +482,153 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+                for filename in filenames:
+                    with zipf.open(filename) as f:
+                        results[filename] = from_json(f.read())
```
timj: Consider stripping the .json from the dictionary here -- the .json is not part of the data model at this point and it makes other areas of the code more model like and less json-specific.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -472,3 +482,153 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+        result = {}
+        with uri.open("rb") as f:
+            with zipfile.ZipFile(f, 'r') as zipf:
+                for filename in zipf.namelist():
+                    with zipf.open(filename) as f:
+                        result[filename] = from_json(f.read())
+        return result
+
+    # Extracting a single file from the ZIP
+    def _extract_blend(self, uri: ResourcePath, blend_filename: str) -> dict[str, Any]:
```
timj: Consider making this the blend_id or blend_ids rather than the explicit filename. The code in here can understand how to go from blend to zip member.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -472,3 +482,153 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+        ----------
+        zip_dict : `dict` of `str` to `object`
+            Dictionary mapping filenames to the desired file type.
+
+        Returns
+        -------
+        model :
+            ScarletModelData instance.
+        """
+        model_psf = zip_dict['psf.json']
```
timj: If you don't mind modifying the dict you can use `zip_dict.pop("psf")` here (assuming .json has been dropped following previous comment) and then it wouldn't even turn up in the loop below.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -472,3 +482,153 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+            'psfShape': psf_shape,
+            'blends': blends,
+        })
+
+    def _model_to_zip(self, model_data: scl.io.ScarletModelData) -> dict[str, Any]:
+        """Convert a ScarletModelData instance to a dictionary of files.
+
+        Parameters
+        ----------
+        model_data : `scl.io.ScarletModelData`
```
timj: Is this in fact `lsst.scarlet.lite.io.ScarletModelData`? Please use the full name since sphinx does not understand import aliases. Similarly it can't work out what `ResourcePath` means on its own.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -472,3 +482,153 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+
+class ScarletModelDelegate(StorageClassDelegate):
+    def can_accept(self, inMemoryDataset: Any) -> bool:
+        return isinstance(inMemoryDataset, scl.io.ScarletModelData)
+
+    def getComponent(self, composite: Any, componentName: str) -> Any:
+        raise AttributeError(f"Unsupported component: {componentName}")
+
+    def handleParameters(self, inMemoryDataset: Any, parameters: Mapping[str, Any] | None = None) -> Any:
+        return inMemoryDataset
```
timj, line 597: Shouldn't this be handling the blend_id parameter and returning a subset?

tests/test_deblend.py
```diff
@@ -81,6 +86,28 @@ def setUp(self):
         for b, coadd in enumerate(self.coadds):
             coadd.setPsf(psfs[b])
 
+        # Initialize a Butler
+        self.root = makeTestTempDir(TESTDIR)
```
timj: I think you can use `makeTestRepo` to simplify things a bit.

timj (commented):
This is great. I'm really happy that the zip approach worked out.  Please document the file format in this package -- looks like there is a doc/lsst.meas.extensions.scarlet/ directory that has a place for it. The documentation is important for requirements verification.

I have one suggestion to unify the code for blend and not blend, and I'm not sure why the delegate doesn't implement the parameter handling. (in-memory butler and InMemoryDatasetHandle testing will use the delegate).

python/lsst/meas/extensions/scarlet/io.py
fred3m: I wasn't sure how this worked/if it was necessary. My thinking was that since fetching the data will have already handled the parameters to load only the selected blends, there is nothing that needs to be done here. Is that assumption wrong?

timj: Yes. Wrong. The delegate is what is used when a file datastore isn't involved. This can be an in-memory datastore or an InMemoryDatasetHandle (that is very useful for testing where people have written tasks that use a deferred dataset handle but don't want to use a butler). This method will be given the main object and then have to deal with the parameters to return the subset.

See for example the delegate for Exposure: https://github.com/lsst/obs_base/blob/main/python/lsst/obs/base/exposureAssembler.py#L280

tests/test_deblend.py
fred3m: I'm not sure how to use this, as I can't figure out where to pass the `run` and it still seems to require creating a temporary root directory and creating an empty Config object. Is there an easier way that I'm missing?

python/lsst/meas/extensions/scarlet/io.py
fred3m: Ah, ok, thanks. Is there a way to test this in a unit test? I mean it's a trivial piece of code but it's still good to be able to test it.

tests/test_deblend.py
timj: Yes. Use pipe base InMemoryDatasetHandle.

python/lsst/meas/extensions/scarlet/io.py
timj: Yes. Use pipe base InMemoryDatasetHandle. (now commenting in the right place)

tests/test_deblend.py
timj: Look in `pipe_tasks/tests/test_getRegionTimeFromVisit.py` for an example. The run can be given to butler.get. You don't need the config because it returns the butler directly.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -472,3 +482,122 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+    """Read and write zip archives.
+
+    In order for files to be read from a zip file, the pydantic model
+    must have a `load` method that accepts a file object and the filename
+    as arguments, as well as a `from_zip` method that accepts a dictionary
+    of files to create the object from the zip archive.
+    """
+
+    default_extension = ".scarlet"
+    unsupported_parameters = None
```
timj: I think you might want this to be `frozenset()` rather than `None`. Are you sure parameters are being passed in and it's not using your storage class delegate to filter after the fact?

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -472,3 +482,122 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+    def can_accept(self, inMemoryDataset: Any) -> bool:
+        return isinstance(inMemoryDataset, scl.io.ScarletModelData)
+
+    def getComponent(self, composite: Any, componentName: str) -> Any:
+        raise AttributeError(f"Unsupported component: {componentName}")
+
+    def handleParameters(self, inMemoryDataset: Any, parameters: Mapping[str, Any] | None = None) -> Any:
+        if "blend_id" in parameters:
+            blend_id = parameters["blend_id"]
+            if isinstance(blend_id, Iterable):
```
timj: you can use `blend_ids = lsst.utils.iteration.ensure_iterable(blend_id)` to clean up this code a bit (and the code above).

PR #109, opened 2025-04-15
DM-50217: Catch failures in spectrum initialization

fred3m:


python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -901,7 +910,7 @@ def deblend(self, mExposure, catalog):
                 t0 = time.monotonic()
                 # Build the parameter lists with the same ordering
                 if self.config.version == "lite":
-                    blend, skippedSources, skippedBands = deblend(
+                    blend, skippedSources, skippedBands, spectrumInit = deblend(
```
arunkannawadi, line 913: Shouldn't the code do something differently if `spectrumInit` is `False`? Or does something else below raise an exception that gets handled?

fred3m: It's not an error. Some blends cannot run the algorithm because they are too big and the node runs out of memory. So in the code below that updates source records it just sets a column in the catalog that notifies us that it skipped spectrum initialization.

arunkannawadi: Looking at the entire code and following along the variable makes sense now.

arunkannawadi (approved):

PR #110, opened 2025-04-24
DM-50486: Read from a local file for the full scarlet model dataset

fred3m:


python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -556,12 +567,9 @@ def read_from_stream(
             filenames = [str(f) for f in filenames]
             filenames += ['psf', 'psf_shape']
         else:
-            filenames = None
+            return NotImplemented("It is inefficient to read the entire archive from a stream.")
```
timj:
```suggestion
            return NotImplemented
```

suggests a problem with test coverage:

```python
>>> NotImplemented("It is inefficient to read the entire archive from a stream.")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'NotImplementedType' object is not callable
```

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -547,6 +548,16 @@ def _model_to_zip(self, model_data: scl.io.ScarletModelData) -> dict[str, Any]:
         })
         return data
 
+    def read_from_local_file(self, path: str, component: str | None = None, expected_size: int = -1) -> Any:
+        # Override of `FormatterV2.read_from_local_file`.
+        with zipfile.ZipFile(path, 'r') as zip_file:
+            unzipped_files = {}
```
timj: I don't like the code duplication here.  Why isn't there a `ScarletModelData.from_stream(stream, blend_ids)` API which is called from `ScarletModelData.from_file(path, blend_ids)` that the formatter calls? How does someone read a file they are sent that is not in butler?

timj (changes requested):
I am concerned that the file reading logic is not in the right place. I realize I should have thought about that last time I reviewed but now that the read logic is in two places it is made more obvious.

python/lsst/meas/extensions/scarlet/io.py
fred3m: Scarlet lite models are typically stored as a single JSON object, so `ScarletModelData` has a `parse_obj` method that extracts all of the model data from JSON into the model data objects. I implemented the zip storage in `meas_extensions_scarlet` and not `scarlet_lite` because I don't want to tie scarlet lite users to any file storage system, so this implementation is only for the science pipelines.

python/lsst/meas/extensions/scarlet/io.py
fred3m: I must not understand how this works. From my understanding https://github.com/lsst/meas_extensions_scarlet/blob/main/tests/test_deblend.py#L386-L389 should be testing this.

fred3m: Also, this ran in ci_hsc, so I'm surprised that it didn't cause an issue there...

python/lsst/meas/extensions/scarlet/io.py
timj: I am concerned that the same data is stored in two different formats but, importantly, if someone passes a file in this .scarlet zip format they can't read it using pipelines code. What's your strategy for how you tell people how to read a file? We can't stop people passing files around outside of butler.

I still don't like the code duplication here though. Please delegate the local file read to the stream reading (and you can factor the stream reading out into a class method if you are not adding a method to ScarletModelData).

fred3m: RE code duplication: I know what you mean, but on the other hand it's only five duplicate lines and the docstring and function signature would require more lines than that. I did think about this but my personal rule is that if creating a new function with docstrings will actually make more lines of code, I just duplicate. But if you feel strongly about it I can change it.

timj: I do feel strongly about this because it's core logic on how to extract the model from the file. Private methods don't need as much documentation but in this case I don't see a problem with a factored out `_read_stream()` private method, with five lines of docstring, that is called directly from the `read_from_stream()` and via `with open() as fd:` call in `read_from_local_file`.

fred3m: Ok, what about this: I can move `_build_model` and `_model_to_zip` from class methods to public functions and also create a `read_scarlet_model` function that reads the file. Then the formatter can call that function from `read_from_local_file` and `read_from_stream`. I can also create a `write_scarlet_model` function that writes to a file and just call that from `ScarletModelFormatter.to_bytes`. Is that ok?

timj: That would work. Still somewhat confused that these aren't methods on `ScarletModelData` itself but public free functions are better than having to tell people to instantiate a formatter to read a file standalone.

fred3m: The main reason I don't include them in scarlet lite is because (as we saw in RFC-1074) there are numerous ways to save hierarchical data, including packages to save zip files. For example, HDF required a different pre-formatting before writing and reading than zip, and ASDF had a different pre-formatting. Given infinite time I'd just implement them all but since we're only implementing zip in the science pipelines it makes more sense to me to keep it in `meas_extensions_scarlet` as opposed to `scarlet_lite`.

fred3m:
Ok, have a look now. I think that I addressed your design concerns although I'm not sure how to improve the testing so that the read_from_local_file method is triggered.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -485,96 +485,141 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
     return blendData
 
 
-class ScarletModelFormatter(TypelessFormatter):
-    """Read and write zip archives.
+def build_scarlet_model(zip_dict: dict[str, Any]) -> scl.io.ScarletModelData:
```
timj, line 488: This doesn't feel like a public function to me. It seems like an implementation detail. Name with leading underscore?

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -485,96 +485,141 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+        'blends': blends,
+    })
+
+
+def read_scarlet_model(path_or_stream: str, blend_ids: list[int] | None = None) -> scl.io.ScarletModelData:
+    """Read a zip file and return a ScarletModelData instance.
+
+    Parameters
+    ----------
+    path : `str`
```
timj, line 518: This is path_or_stream and so needs also a different type.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -485,96 +485,141 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+    for key, value in zip_dict.items():
+        blends[int(key)] = value
+    return scl.io.ScarletModelData.parse_obj({
+        'psf': model_psf,
+        'psfShape': psf_shape,
+        'blends': blends,
+    })
+
+
+def read_scarlet_model(path_or_stream: str, blend_ids: list[int] | None = None) -> scl.io.ScarletModelData:
```
timj, line 513: Wrong type annotation? Also, I think you likely want to add this to `__all__`.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -485,96 +485,141 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+    })
+    return data
+
+
+def write_scarlet_model(path_or_stream: str | BinaryIO, model_data: scl.io.ScarletModelData):
+    """Write a ScarletModelData instance to a zip file.
+
+    Parameters
+    ----------
+    model_data : `lsst.scarlet.lite.io.ScarletModelData`
```
timj, line 581: Missing a parameter.

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -485,96 +485,141 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+        for blend_id, blend_data in json_model['blends'].items()
+    }
+    data.update({
+        'psf_shape': json.dumps(json_model['psfShape']),
+        'psf': json.dumps(json_model['psf']),
+    })
+    return data
+
+
+def write_scarlet_model(path_or_stream: str | BinaryIO, model_data: scl.io.ScarletModelData):
```
timj, line 576: Add to `__all__`?

python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -485,96 +485,141 @@ def oldScarletToData(blend: Blend, psfCenter: tuple[int, int], xy0: Point2I):
...
+        if filenames is None:
+            filenames = zip_file.namelist()
+        for filename in filenames:
+            with zip_file.open(filename) as f:
+                unzipped_files[filename] = from_json(f.read())
+
+        return build_scarlet_model(unzipped_files)
+
+
+def scarlet_model_to_zip_json(model_data: scl.io.ScarletModelData) -> dict[str, Any]:
```
timj, line 547: Private implementation detail so use underscore?

timj (approved):
Looks good. Please squash the first two commits together.

PR #112, opened 2025-09-08
DM-52299: Properly handle missing PSF information in coadds with little to no data

fred3m:


python/lsst/meas/extensions/scarlet/utils.py
```diff
@@ -227,15 +228,18 @@ def computeNearestPsfMultiBand(
     for band in mExposure.filters:
         psf, psfCenter, diff = computeNearestPsf(
             mExposure[band,],
-            psfCenter,
             catalog,
             band,
+            psfCenter,
```
kherner, line 233: Is it correct to switch the order here? I haven't seen the computeNearestPsf code.

kherner (approved):
Just one minor question.

python/lsst/meas/extensions/scarlet/utils.py
fred3m: Yes, it is necessary here. This code block wasn't used before and the previous implementation was a syntax error. Here's signature of `computeNearestPsf`: https://github.com/lsst/meas_extensions_scarlet/blob/main/python/lsst/meas/extensions/scarlet/utils.py#L124-L129

fred3m: I only found this because this PR hit this code block and raised an Exception until I made the change.

PR #113, opened 2025-09-18
DM-52555: Raise NoWorkFound when scarlet model is missing a band

fred3m:


python/lsst/meas/extensions/scarlet/io.py
```diff
@@ -174,7 +174,12 @@ def updateCatalogFootprints(
     if modelData.metadata is None:
         raise ValueError("Scarlet model data does not contain metadata")
     bands = modelData.metadata["bands"]
-    bandIndex = bands.index(band)
+    try:
+        bandIndex = bands.index(band)
+    except ValueError:
+        # Import pipe.base here to avoid a circular import
```
erykoff: How can this be a circular import?  E.g., how could lsst.pipe.base depend on scarlet?

fred3m: You're right, I was thinking it was importing pipe_tasks, not pipe_base. In fact, it already depends on pipe_base so this is fine. I'll import normally.

erykoff (approved):

PR #114, opened 2025-09-22
DM-51672: Refactor tests

fred3m:


python/lsst/meas/extensions/scarlet/utils.py
```diff
@@ -125,8 +127,8 @@ def computePsfKernelImage(mExposure, psfCenter, catalog=None):
 def computeNearestPsf(
     calexp: Exposure,
     catalog: SourceCatalog,
-    band: str,
-    psfCenter: Point2D,
+    band: str | None = None,
```
taranu, line 130: Clarify what `band=None` does in the docstring.

tests/test_deblend.py
```diff
@@ -28,154 +28,235 @@
...
             deconvolvedCoadds.append(deconvolvedCoadd)
         mDeconvolved = afwImage.MultibandExposure.fromExposures(self.bands, deconvolvedCoadds)
+        return mDeconvolved
+
+    def test_default_deconvolve(self):
+        data = self.initialize_data(self.models)
+        deconvolved = self.deconvolve(data)
+
+        diff = data.deconvolved.data - deconvolved.image.array
+        # Due to peakiness of Sersic models the center has a sharp peak,
```
taranu, line 215: They're all exponential, no? That's not especially peaky...

tests/test_deblend.py
```diff
@@ -359,12 +438,37 @@ def _test_blend(self, blendData1, blendData2, model_psf, psf, bands):
         )
         np.testing.assert_almost_equal(blend1.get_model().data, blend2.get_model().data)
 
+    def _setup_butler(self):
```
taranu: I suggest moving this to the end of the class' methods rather than having it sandwiched between tests.

tests/test_deblend.py
```diff
@@ -437,6 +542,206 @@ def test_legacy_model(self):
...
+
+    def _generateGoodPsf(self, sigma: float = 1.0):
+        # Generate a PSF and Image of the PSF
+        psfRadius = 20
+        psfShape = (2 * psfRadius + 1, 2 * psfRadius + 1)
+        psf = GaussianPsf(psfShape[1], psfShape[0], sigma)
+        psfImage = psf.computeImage(psf.getAveragePosition()).array
+        return psf, psfImage
+
+    def _generateMultibandPsf(self, sigmas: list[float]):
```
taranu: Same comment here about putting private methods at the end instead of between tests.

tests/test_deblend.py
```diff
@@ -437,6 +542,206 @@ def test_legacy_model(self):
...
+                footprint.addPeak(*peak)
+            src.setFootprint(footprint)
+
+            for band in bands:
+                src[f"merge_footprint_{band}"] = True
+                footprint.peaks[f"merge_peak_{band}"] = True
+        return catalog
+
+    def test_computeNearestPsfGood(self):
+        # Test that using a valid PSF works noramlly
```
taranu: normally

tests/test_deblend.py
```diff
@@ -437,6 +542,206 @@ def test_legacy_model(self):
...
+        coadd = self._generateCoadd(psf)
+
+        # Test that computing the PSF works
+        derivedPsf, center, dist = mes.utils.computeNearestPsf(coadd, None, "g", Point2D(25, 25))
+        np.testing.assert_array_equal(derivedPsf.array, psfImage)
+        self.assertEqual(center, Point2D(25, 25))
+        self.assertEqual(dist, 0)
+
+    def test_computeNearestPsfRecoverable(self):
+        # Test that using a PSF not defined at the inital location
```
taranu: initial

tests/test_deblend.py
```diff
@@ -437,6 +542,206 @@ def test_legacy_model(self):
...
+        # will fallback to a valid location.
+        psf, psfImage = self._generateGoodPsf()
+        coadd = self._generateCoadd(BadPsf(Point2D(1, 1), psf))
+        catalog = self._generateCatalog(self.bands, [[(1, 1, 10)]])
+
+        # Test that computing the PSF works after finding a new location
```
taranu: Maybe clarify that `new location` means the peak you added at `(1, 1)` in the previous line (this took me a bit to work out).

tests/test_deblend.py
```diff
@@ -437,6 +542,206 @@ def test_legacy_model(self):
...
+
+    def test_computeNearestPsfMultiBandRecoverable(self):
+        # Test that a Psf at a different location is still recoverable
+        bands = tuple("gri")
+        psfs, psfImage = self._generateMultibandPsf([1.0, 1.2, 1.4])
+        psfs[1] = BadPsf(Point2D(1, 1), psfs[1])
+        mCoadd = self._generateMultibandCoadd(psfs, bands)
+        catalog = self._generateCatalog(self.bands, [[(1, 1, 10)]])
+
+        # Test that computing the PSF works
```
taranu: (because the catalog has a peak at the location of the BadPsf)

tests/test_deblend.py
```diff
@@ -437,6 +542,206 @@ def test_legacy_model(self):
...
+    def test_computeNearestPsfMultiBandIncomplete(self):
+        # Test that missing a PSF in one band returns a PSF and
+        # an exposure that are missing bands.
+        bands = tuple("gri")
+        psfs, psfImage = self._generateMultibandPsf([1.0, 1.2, 1.4])
+        psfs[1] = BadPsf(Point2D(1, 1), psfs[1])
+        mCoadd = self._generateMultibandCoadd(psfs, bands)
+        catalog = self._generateCatalog(self.bands)
+
+        # Test that computing the PSF works
```
taranu: (for the non-bad g & i PSFs)

tests/test_deblend.py
```diff
@@ -437,6 +542,206 @@ def test_legacy_model(self):
...
+        psfs, psfImage = self._generateMultibandPsf([1.0, 1.2, 1.4])
+        psfs = [BadPsf(Point2D(1, 1), psfs) for psf in psfs]
+        mCoadd = self._generateMultibandCoadd(psfs, bands)
+        catalog = self._generateCatalog(self.bands)
+
+        with self.assertRaises(NoWorkFound):
+            mes.utils.buildObservation(modelPsf, Point2I(25, 25), mCoadd, catalog=catalog)
+
+    def test_buildObservationBadPsfs(self):
+        # Test that creating an observation with all bad PSFs
```
taranu: Couldn't you just add:
```
        with self.assertRaises(NoWorkFound):
            mes.utils.buildObservation(modelPsf, Point2I(25, 25), mCoadd)
```
... to the test above instead of copying everything else?

tests/utils.py
```diff
@@ -19,81 +19,95 @@
...
-            [
-                scipy.signal.convolve(m, targetPsfImage, method="direct", mode="same")
-                for m in morphs
-            ]
-        )
-        morphs /= morphs.max()
-        psfImages /= psfImages.sum(axis=(1, 2))[:, None, None]
+    for dim in bbox.origin:
+        if dim < 0:
+            raise ValueError("Invalid setup, at least one source has a footprint below (0, 0)")
```
taranu, line 106: Why only check for the origin being below `(0, 0)` rather than all corners being within the scene?

taranu (commented):
Some comments, questions and suggestions as usual.

tests/test_deblend.py
fred3m: They're more peaky than real galaxies because scarlet lite does not fit these centers very well (compared to the rest of the galaxy) but has no problem with actual galaxies.

tests/utils.py
fred3m: This is due to the way that python indexing and slicing works. When Peter and I wrote scarlet Box and when I wrote the scarlet lite Image class we elected to use ordinary python index slicing rather than allow them to treat negative indices as coordinates below 0. So having boxes outside of the positive bounds of the image is fine, they just can't have negative indices.

PR #116, opened 2025-10-17
DM-52564: Implement IsolatedSources

fred3m:


python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -0,0 +1,156 @@
...
+        f"to match the ScarletModelData schema version {scl.io.model_data.CURRENT_SCHEMA}."
+    )
+
+
+class LsstModelData(scl.io.ScarletModelData):
+    """A ScarletModelData that includes isolated sources.
+
+    Attributes
+    ----------
+    isolated :
```
TallJimbo: Does this style with the types omitted work?

(I'm hoping it will when we have a new Sphinx that can read type annotations, but I don't think we're there yet.)

python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -0,0 +1,156 @@
...
+        dtype :
+            Datatype of the resulting model.
+
+        Returns
+        -------
+        result :
+            The reconstructed object
+        """
+        data = scl.io.migration.MigrationRegistry.migrate(MODEL_TYPE, data)
+        blends: dict[int, scl.io.ScarletBlendBaseData] | None = {}
```
TallJimbo: Looks like this isn't actually ever `None`?

python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -0,0 +1,156 @@
...
+        metadata = data.get("metadata", None)
+        for bid, blend in data.get("blends", {}).items():
+            if "blend_type" not in blend:
+                # Assume that this is a legacy model
+                blend["blend_type"] = "blend"
+            try:
+                blend_data = scl.io.ScarletBlendBaseData.from_dict(blend, dtype=dtype)
+            except KeyError:
+                raise ValueError(f"Unknown blend type: {blend['blend_type']} for blend ID: {bid}")
+            blends[int(bid)] = blend_data  # type: ignore
```
TallJimbo: Given what `as_dict` looks like, I'm surprised you can't delegate more of the above to the base class.  Is the backwards-compatibility such that it has to happen here rather than in scarlet lite?

python/lsst/meas/extensions/scarlet/io/source_data.py
```diff
@@ -0,0 +1,132 @@
...
+        Returns
+        -------
+        result :
+            The object encoded as a JSON-compatible dictionary.
+        """
+        result: dict[str, Any] = {
+            "origin": tuple(int(o) for o in self.origin),
+            "shape": tuple(int(s) for s in self.footprint.shape),
+            "peak": tuple(float(p) for p in self.peak),
+            "footprint": tuple(self.footprint.flatten().astype(float)),
```
TallJimbo: I gather this isn't actually a footprint; it's really more like a postage stamp, with the area outside the footprint set to zero.  Might be better to call it something else.

python/lsst/meas/extensions/scarlet/io/source_data.py
```diff
@@ -0,0 +1,132 @@
...
+        ----------
+        data :
+            Dictionary representation of the object
+        dtype :
+            Datatype of the resulting model.
+
+        Returns
+        -------
+        result :
+            The reconstructed object
```
TallJimbo:
```suggestion
            The reconstructed object.
```

TallJimbo (approved):

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: Sigh. No, this is just a habit from my other projects because Sphinx does automatically use type annotations if they're available, and apparently so does documenteer, but not the versions that we're using. ☹️

python/lsst/meas/extensions/scarlet/io/source_data.py
fred3m: Changed to `span_array`.

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: I just had to make a couple of small changes to scarlet_lite. But now the creation of blends is done in `scarlet_lite` and only isolated sources is done here.

PR #120, opened 2025-11-04
DM-53142: Fix bug in isolated source persistence

fred3m:


taranu (commented):
The additions look fine, but how are you testing that backwards compatibility is maintained? I don't see any new tests related to that.

fred3m:
That test already exists: https://github.com/lsst/meas_extensions_scarlet/blob/main/tests/test_deblend.py#L512-L561. That directory will hold legacy models so that they can be tested with each new release.

taranu:
> That test already exists: https://github.com/lsst/meas_extensions_scarlet/blob/main/tests/test_deblend.py#L512-L561. That directory will hold legacy models so that they can be tested with each new release.

That seems to cover v29 models. Does that ensure that this new version will correctly load w_2025_44 and w_2025_45 models (assuming it's merged today)? As in, it will implicitly cover both branches of the `blend_type`/`source_type` conditional?

tests/test_deblend.py
```diff
@@ -455,6 +455,15 @@ def test_persistence(self):
...
 
+        for sourceId in modelData.isolated.keys():
+            isolatedData1 = modelData.isolated[sourceId]
+            isolatedData2 = modelData2.isolated[sourceId]
+            self.assertTupleEqual(isolatedData1.origin, isolatedData2.origin)
+            np.testing.assert_array_equal(
+                isolatedData1.span_array,
+                isolatedData2.span_array,
+            )
+
```
fred3m, line 466: There are no changes in the `w44` and `w45` models. This PR fixed an error where the current models were not being loaded correctly from disk (I had forgotten how the `ScarletModelFormatter` worked by saving the blends to zip files and I never extracted the isolated objects).

The previous PR tested that the isolated sources existed when it ran the deblender but didn't test that they were persisted. This code block in the new PR is where I test that the isolated sources were saved and loaded correctly. Had I put this in the previous PR it would have caught this error.

PR #122, opened 2025-12-09
DM-53388: Add copy, deepcopy, and getitem to IsolatedSource

fred3m:


tests/test_isolated_source.py
```diff
@@ -0,0 +1,153 @@
...
+class ScarletTestCase(TestCase):
+    """A base TestCase for scarlet tests.
+    """
+    def setUp(self) -> None:
+        super().setUp()
+        self.bands = tuple("grizy")
+        peak = (27, 32)
+        bbox = scl.Box((15, 15), (20, 25))
+        morph = scl.utils.integrated_circular_gaussian(sigma=0.8).astype(np.float32)
+        spectrum = np.arange(5).astype(np.float32)
```
taranu: You can do `np.arange(5, dtype=np.float32)`

tests/test_isolated_source.py
```diff
@@ -0,0 +1,153 @@
...
+        model_image = scl.Image(model, yx0=bbox.origin, bands=self.bands)
+        self.component = scl.component.CubeComponent(model=model_image, peak=peak)
+
+    def test_constructor(self):
+        source = mes.source.IsolatedSource(
+            model=self.component._model,
+            peak=self.component.peak,
+        )
+
+        np.testing.assert_array_equal(
```
taranu, line 50: The args aren't copied, are they? If so, isn't testing array equality unnecessary compared to since `source.component._model is self.component._model` should be True?

tests/test_isolated_source.py
```diff
@@ -0,0 +1,153 @@
...
+            source.component._model.data,
+            self.component._model.data,
+        )
+        self.assertEqual(source.component.peak, self.component.peak)
```
taranu, line 54: Similar comment here, except I guess it tests `peak.__eq__`?

tests/test_isolated_source.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def test_copy(self):
+        source = mes.source.IsolatedSource(
+            model=self.component._model,
+            peak=self.component.peak,
+        )
+        source_copy = source.copy()
+
+        self.assertIsNot(source_copy, source)
+        np.testing.assert_array_equal(
```
taranu: Seems like these should be `assertIs` as well for a shallow copy.

tests/test_isolated_source.py
```diff
@@ -0,0 +1,153 @@
...
+        self.assertEqual(source_copy.component.peak, source.component.peak)
+
+    def test_deep_copy(self):
+        source = mes.source.IsolatedSource(
+            model=self.component._model,
+            peak=self.component.peak,
+        )
+        source_copy = source.copy(deep=True)
+
+        self.assertTupleEqual(source_copy.component.peak, source.component.peak)
```
taranu, line 77: Should there also be some `assertIsNot` here?

tests/test_isolated_source.py
```diff
@@ -0,0 +1,153 @@
...
+        source = mes.source.IsolatedSource(self.component._model, self.component.peak)
+        indices = ("i", "g", "r")
+        source_reordered = source[indices]
+        self.assertTupleEqual(source_reordered.bands, indices)
+        np.testing.assert_array_equal(
+            source_reordered.get_model().data,
+            source.get_model().data[[2, 0, 1]],
+        )
+
+        source_reordered = source["igr"]
```
taranu, line 114: This is relying on string iterability, right? I would try to avoid that in production code.

tests/test_isolated_source.py
```diff
@@ -0,0 +1,153 @@
...
+        with self.assertRaises(IndexError):
+            source[:, 10:20, 10:20]
+
+        with self.assertRaises(IndexError):
+            source[1:]
+
+        with self.assertRaises(IndexError):
+            source[1]
+
+        with self.assertRaises(IndexError):
```
taranu, line 164: Is this because of the box origin, and `source[15, 15]` would work? You may want to make a note of that.

tests/test_isolated_source.py
fred3m: Yes, this is just verifying that the model is being initialized and stored properly. Obviously it is now but I want to make sure that a bad refactor or something else in the future doesn't change that.

tests/test_isolated_source.py
fred3m: Yes. I'm of the opinion that it's better to test the obvious than not test something and have a later breakage.

tests/test_isolated_source.py
fred3m: We do not use this (or anything else in this PR) in production code but this is how selecting on filters has worked in scarlet lite images since the beginning. This is just ensuring that the same indexing scheme for blends and sources in `meas_extensions_scarlet` is the same.

tests/test_isolated_source.py
fred3m: No, that wouldn't work either. Just like the afw Image class the user is required to using a bounding box for spatial dimensions to avoid confusion as to whether or not an origin is being applied.

fred3m: Agreed about documentation in the tests.Once this is merged, unless I'm given some more random bug fix tickets like I was at the end of 2025, I also plan on writing docs and updating the deblending tutorials.

PR #124, opened 2025-12-19
DM-53685: Make deconvolution more robust against non-finite values

fred3m:


python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
```diff
@@ -265,7 +268,13 @@ def _buildObservation(
         bands = (band,)
         model_psf = scl.utils.integrated_circular_gaussian(sigma=0.8)
 
+        # Give zero weight to non-finite pixels
+        weights = np.ones_like(coadd.image.array)
+        weights[~np.isfinite(coadd.image.array)] = 0
+
         image = coadd.image.array
+        # Set non-finite pixels to zero
+        image[~np.isfinite(image)] = 0.0
```
erykoff, line 287: Does this modify the image in place?  If so, is that a problem?

fred3m: That's true, I'll make a copy first.

arunkannawadi: I thought it is already a deep copy of the input image.

arunkannawadi: https://github.com/lsst/meas_extensions_scarlet/blob/978e50f9ea318bc3a5399712dc0a836b3c9b823d/python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py#L295

fred3m: I think that you're right if it's a cell coadd from `runQuantum` but in general a user could call this with their own exposure and this would have updated it in place before (or maybe you didn't see it until I pushed my change).

fred3m: Ah, now that I'm copying the image to remove the NaNs I don't need to copy again.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
```diff
@@ -60,6 +60,9 @@ def calculate_update_step(observation: scl.Observation) -> float:
     """
     # Calculate sparsity as fraction of pixels significantly above noise
     noise_level = observation.noise_rms[0]
+    # Guard against non-finite or non-positive noise levels
+    if noise_level <= 0 or not np.isfinite(noise_level):
+        return 0.1
```
arunkannawadi: Should we have this as an optional argument to this function, or are there enough magic numbers in here already>

fred3m: It doesn't really matter. This is just a convenience function that speeds up conversion a bit, but for good coding practice I did modify this by adding a couple of parameters and replacing the magic numbers.

arunkannawadi (approved):

fred3m:
I added a similar fix in `buildObservation` for the deblender so that NaN values in observations will also not affect deblending. We must not have run into that problem because in this case the NaN wasn't in a blend, but it still seems like a good practice.

PR #127, opened 2026-02-02
DM-53888: Make initialization more robust against NaN values

fred3m:


python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -152,8 +152,8 @@ def _getDeconvolvedFootprints(
     """
     bbox = mDeconvolved.getBBox()
     xmin, ymin = bbox.getMin()
-    sigma = np.median(np.sqrt(mDeconvolved.variance.array), axis=(1, 2))
-    detect = np.sum(mDeconvolved.image.array/sigma[:, None, None], axis=0)
+    sigma = np.nanmedian(np.sqrt(mDeconvolved.variance.array), axis=(1, 2))
+    detect = np.nansum(mDeconvolved.image.array/sigma[:, None, None], axis=0)
```
arunkannawadi, line 156: The first change alone should suffice, right?

arunkannawadi (approved):
LGTM! There are many more NaN-unsafe aggregates in this module that might be worth switching, but not necessarily in a ticket intended specifically for backporting

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
fred3m: Yes, but I figure that there is no harm in protecting against NaN values in the image as well. I actually found one w45 case where there was a NaN in one image band so I added that `nansum` just to ensure that it fixed the problem on real data from DM-53885.

arunkannawadi: Yeah, reminds me that one time where there was a pesky NaN in one of the cell coadds.

PR #130, opened 2026-06-01
DM-54841: Fix bugs found in audit

fred3m:


natelust:
I did not look line by line at code line change. I did try to get the meaning of each commit at it seemed reasonable (also I am not a scarlet expert). There was one or two things that caught my eye to look at. Also at one point the meanings of columns seems to change in catalogs. This might be a problem, even if what it is now is more correct, as people might be developing around what it is. Take a look though you are the expert.

One thing to note. LLM are really good at being able to spam out unit tests, which is good for preserving behavior. I am a bit worried about the quantity of them though, it does take time to spin up runners for all these tests, and we we get too many it may start to harm build install and testing times. It might make sense to consolidate many of the tests into larger single method tests that have several different asserts in them and group by larger concepts. Not really required here specifically, but something to keep in mind that might be worth doing now or later.

fred3m:
Yes, there was a difference in the documented behavior of a feature and the implementation of it. Either way was going to affect downstream users and I made the decision to stay true to the documented (intended behavior) with a note in the release notes to notify users. I justify this because the behavior change now matches *the name of the flag*. If I didn't change the behavior then the actual names of the flags were misleading, so I think that in this case the sign flip was justifiable.

Re tests: I agree and I actually put a lot of thought into this. This is why the test refactor cached images, deblender outputs, etc. Before the refactor, my machine tool 5.56s to run 24 tests. Now it takes 8.17s to run 128 tests. The key is that the parts that take time to generate are cached and re-used in all of the subtests. This enabled me to have a lot more specific tests to significantly increase the code coverage and catch all of the bugs that existed before, and a few new ones.

PR #131, opened 2026-06-09
DM-55109: Create lsst dataclasses for some persistable objects

fred3m:
The science pipelines were abusing the use of `metadata` fields in scarlet lite dataclasses for persistence, making changes that could not be properly migrated. This ticket makes the models used in the LSST science pipelines more robust and also refactored the legacy tests to factor out some of the boiler plate.


python/lsst/meas/extensions/scarlet/io/utils.py
```diff
@@ -762,14 +770,14 @@ def loadBlend(
...
-                "`modelData.metadata` must be populated to use the "
-                "`modelData` branch of `loadBlend`."
+                "`modelData` must carry `bands`, `psf` and `model_psf` to use "
+                "the `modelData` branch of `loadBlend`."
             )
-        bands = tuple(modelData.metadata["bands"])
-        psfs = modelData.metadata["psf"]
-        actual_model_psf = modelData.metadata["model_psf"][None, :, :]
+        bands = tuple(modelData.bands)
+        psfs = modelData.psf
```
taranu, line 779: I suppose it's too late to change but it seems a bit of a shame that `Observation` uses `psfs` while `modelData` and other classes use just `psf` for the same thing.

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
```diff
@@ -0,0 +1,246 @@
...
+
+from .source_data import _decode_span_array, _encode_span_array
+
+__all__ = ["LsstHierarchicalBlendData"]
+
+CURRENT_SCHEMA = "1.0.0"
+BLEND_TYPE = "lsst_hierarchical"
+scl.io.migration.MigrationRegistry.set_current(BLEND_TYPE, CURRENT_SCHEMA)
+
+# The legacy scarlet_lite ``blend_type`` tags whose blends carry the
```
taranu: Does this comment get parsed with formatting in the docs? If so, does BLEND_TYPE need a quasi-docstring too?

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
```diff
@@ -0,0 +1,246 @@
...
+    version: str = CURRENT_SCHEMA
+    children: dict[int, scl.io.ScarletBlendBaseData]
+    span_array: np.ndarray
+    origin: tuple[int, int]
+    legacy_spans: bool = False
+
+    @property
+    def shape(self) -> tuple[int, int]:
+        """The ``(height, width)`` of the footprint span mask."""
+        return tuple(int(s) for s in self.span_array.shape)  # type: ignore[return-value]
```
taranu: Why is the ignore necessary?

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
```diff
@@ -0,0 +1,246 @@
...
+            except KeyError:
+                raise scl.io.utils.PersistenceError(
+                    f"Unknown blend type: {child.get('blend_type')} for blend ID: {blend_id}"
+                )
+        shape = tuple(int(s) for s in data["shape"])
+        # Spans are a boolean detection mask; the model ``dtype`` does not
+        # apply to them.
+        span_array = _decode_span_array(data["span_array"], shape, bool)
+        origin = tuple(int(o) for o in data["origin"])
+        legacy_spans = bool(data.get("legacy_spans", False))
```
taranu, line 138: I don't see how it would be harmful but why is the cast needed?

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
```diff
@@ -0,0 +1,246 @@
...
+            legacy_spans=legacy_spans,
+            metadata=metadata,
+        )
+
+    @classmethod
+    def convert_from_hierarchical(cls, blend_dict: dict) -> dict:
+        """Promote a legacy scarlet_lite ``hierarchical`` blend dict to the
+        ``lsst_hierarchical`` shape.
+
+        Reads the *frozen* legacy on-disk hierarchical format directly: a
```
taranu: This is a confusing way to describe it. It's taking a dict as input, so clearly not reading an on-disk format directly but converting from an old format to new.

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
```diff
@@ -0,0 +1,246 @@
...
+        ``children`` map plus the meas-specific ``spans``/``origin`` that the
+        deblender wrote into the blend ``metadata`` dict. It does NOT route the
+        input through scarlet_lite's live ``hierarchical`` migration chain, so
+        future changes to ``HierarchicalBlendData`` do not affect this
```
taranu: I don't see how this information is useful to users.

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
```diff
@@ -0,0 +1,246 @@
...
+        one-time legacy promotion. The ``children`` dicts are passed through
+        untouched and are migrated individually when they are later
+        deserialized by :meth:`from_dict`.
+
+        Real spans stored in ``metadata["spans"]`` are promoted to a typed
+        ``span_array`` (``legacy_spans=False``). When no spans are present
+        (truly ancient archives) a filled rectangle covering the children's
```
taranu: A version number of some kind would be more useful here.

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
```diff
@@ -0,0 +1,246 @@
...
+        combined bounding box is synthesized and ``legacy_spans=True`` is set
+        — a reconstruction, not the original detection footprint.
```
taranu: Excessive verbiage. It could just say something like `If no spans are present, calls `_bbox_from_children` to return a superset of the child bboxes`. The docstring for that method should cover the details.

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
```diff
@@ -0,0 +1,246 @@
...
+        }
+        if residual:
+            result["metadata"] = scl.io.utils.encode_metadata(residual)
+        return result
+
+    @staticmethod
+    def _bbox_from_children(children: dict) -> tuple[tuple[int, int], tuple[int, int]]:
+        """Compute the ``(origin, shape)`` covering all child blend dicts.
+
+        Used only by the legacy-spans synthesis path. Children are the flat
```
taranu: I don't think it's a good idea to declare where a method is currently used as that places a burden on updating the doc if it's ever called elsewhere. Better say `Should only be used by...` if that's true, or `Needed by...` if not, and if that's actually useful context.

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
```diff
@@ -0,0 +1,246 @@
...
+        `~lsst.scarlet.lite.io.ScarletBlendData` dicts produced by the
+        deblender, each carrying ``origin``/``shape``. Nested hierarchical
```
taranu: This part would be better off as separate entries in a `Raises` section.

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
```diff
@@ -0,0 +1,246 @@
...
+        shapes = []
+        for child in children.values():
+            if "origin" not in child or "shape" not in child:
+                raise NotImplementedError(
+                    "Legacy span synthesis only supports flat child blends with "
+                    f"'origin'/'shape'; got blend_type {child.get('blend_type')!r}."
+                )
+            origins.append(tuple(int(o) for o in child["origin"]))
+            shapes.append(tuple(int(s) for s in child["shape"]))
+        min_y = min(o[0] for o in origins)
```
taranu, line 229: Iterating over origins four times seems wasteful but I suppose this is not a performance bottleneck.

python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -21,82 +21,84 @@
...
-
-
-# Ensure that the ScarletModelData from scarlet lite hasn't changed.
-_checkScarletLiteSchema(scl.io.model_data.CURRENT_SCHEMA, SCARLET_LITE_SCHEMA)
-
-
-class LsstScarletModelData(scl.io.ScarletModelData):
-    """A ScarletModelData that includes isolated sources.
+    This mirrors `~scarlet_lite.io.ScarletModelData` except that it includes
+    information speficific to the LSST science pipelines, which has
```
taranu: specific

A brief summary of those differences could be useful here.

python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -21,82 +21,84 @@
...
     Attributes
     ----------
-    isolated : dict[int, IsolatedSourceData]
+    blends
+        A mapping of parent IDs to blend data.
+    isolated
         A mapping of isolated source IDs to their data.
-    version : dict[int, scl.io.ScarletBlendBaseData]
+    metadata
+        Residual metadata transport (load-bearing fields are lifted out).
```
taranu: This is not a useful description of what's actually in it now, and the implied context that it used to contain more items is not helpful here.

python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -21,82 +21,84 @@
...
         metadata: dict[str, Any] | None = None,
+        bands: tuple[str, ...] | None = None,
+        model_psf: np.ndarray | None = None,
+        psf: np.ndarray | None = None,
     ):
-        super().__init__(blends=blends, metadata=metadata)
+        self.blends = blends if blends is not None else {}
+        self.metadata = metadata
         self.isolated = isolated if isolated is not None else {}
+        # ``metadata`` is transport only (see class docstring). Lift the
```
taranu: Again this is way too many words to say that important keys are moved out of the dict into attributes. What does `in-memory contract` even mean? They're in memory either way.

python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -21,82 +21,84 @@
...
+        # load-bearing fields into typed attributes (the in-memory
+        # contract), preferring explicit constructor args and otherwise
+        # pulling them out of the transported metadata. They are removed
+        # from the in-memory metadata so the attributes are the single
+        # source of truth; ``as_dict`` writes them back into the serialized
+        # metadata blob.
+        meta = self.metadata if self.metadata is not None else {}
+        self.bands = (
```
taranu: Should there be validation that `bands` is _not_ in `meta` if it's provided as a kwarg, or is it allowed to specify both? If the latter, then the kwarg taking priority should probably be documented.

python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -21,82 +21,84 @@
...
+            tuple(bands) if bands is not None
+            else (tuple(meta["bands"]) if meta.get("bands") is not None else None)
+        )
+        self.model_psf = model_psf if model_psf is not None else meta.get("model_psf")
```
taranu: Same question here about whether there should be validation of potentially conflicting kwarg/metadata values.

python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -106,15 +108,26 @@ def as_dict(self) -> dict[str, Any]:
...
-        data.update(
-            {
-                "model_type": MODEL_TYPE,
-                "isolated": {k: v.as_dict() for k, v in self.isolated.items()},
-                "version": self.version,
-            }
-        )
-        return data
+        # Reassemble the transport metadata blob from the typed load-bearing
+        # attributes (kept metadata-resident on disk; see class docstring).
```
taranu: What does `metadata-resident on disk` even mean?

python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -106,15 +108,26 @@ def as_dict(self) -> dict[str, Any]:
...
+        meta = dict(self.metadata) if self.metadata is not None else {}
+        if self.bands is not None:
+            meta["bands"] = tuple(self.bands)
+        if self.model_psf is not None:
+            meta["model_psf"] = self.model_psf
+        if self.psf is not None:
+            meta["psf"] = self.psf
+        return {
```
taranu, line 163: If these keys are unlikely to change much, perhaps they should be expanded in the `result` docstring.

python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -198,3 +226,42 @@ def _to_1_0_1(data: dict) -> dict:
...
+
+    1.0.1 (and earlier) stored top-level parent blends as scarlet_lite
+    ``HierarchicalBlendData`` with the detected-parent footprint hidden in the
+    blend ``metadata`` dict (``spans``/``origin``) -- a load-bearing,
+    cross-package field that neither package could version cleanly. 1.0.2
+    promotes each such blend to ``LsstHierarchicalBlendData``
+    (``blend_type="lsst_hierarchical"``), where those fields are typed,
+    version-controlled attributes.
+
+    The model migration only *orchestrates* the promotion: each legacy
```
taranu: This paragraph is longer than the actual code and doesn't say much more than that each blend is converted by calling `convert_from_hierarchical`.

If you want to add some (more?) documentation on the differences between 1.0.1 and 1.0.2 model schema then I don't think the docstring of a single-under method is the best place for that.

python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -198,3 +226,42 @@ def _to_1_0_1(data: dict) -> dict:
...
+    data : dict
+        The data to migrate.
+    Returns
+    -------
+    result : dict
+        The migrated data.
+    """
+    blends = data.get("blends", {})
+    for bid, blend in list(blends.items()):
+        if blend.get("blend_type", "blend") in LEGACY_HIERARCHICAL_TYPES:
```
taranu, line 310: Should this raise if `blend_type` is not `BLEND_TYPE`?

python/lsst/meas/extensions/scarlet/io/utils.py
```diff
@@ -207,22 +207,31 @@ def updateCatalogFootprints(
         # to do in this function. This is rare but it does occasionally
         # happen in fields that only have u-band images.
         return
-    if modelData.metadata is None:
-        raise ValueError("Scarlet model data does not contain metadata")
-    bands = modelData.metadata["bands"]
+    if modelData.bands is None or modelData.model_psf is None or modelData.psf is None:
+        raise ValueError(
+            "Scarlet model data is missing the `bands`, `model_psf` or `psf` "
+            "needed to update catalog footprints."
```
taranu, line 213: Is this something users running outside of a pipetask are likely to encounter? If so, would it help to suggest trying to migrate legacy data if that's the problem?

tests/test_io_hierarchical_blend_data.py
```diff
@@ -0,0 +1,138 @@
...
+        preserves the provenance marker and bounding box.
+        """
+        spans = _spans()
+        blend = LsstHierarchicalBlendData(
+            children={1: _child()}, span_array=spans, origin=(10, 20)
+        )
+        blend2 = ScarletBlendBaseData.from_dict(blend.as_dict())
+        self.assertIsInstance(blend2, LsstHierarchicalBlendData)
+        np.testing.assert_array_equal(blend2.span_array, spans)
+        self.assertEqual(tuple(blend2.origin), (10, 20))
```
taranu: Why not just compare all of the attributes in a loop against the original blend values instead of picking and choosing a few?

tests/test_io_hierarchical_blend_data.py
```diff
@@ -0,0 +1,138 @@
...
+    spans[1:4, 2:5] = True
+    return spans
+
+
+def _child():
+    return scl.io.ScarletBlendData(origin=(10, 20), shape=(5, 6), sources={})
+
+
+class TestLsstHierarchicalBlendData(lsst.utils.tests.TestCase):
+    """Round-trip, conversion, and lineage tests for the new blend class."""
```
taranu: I don't think it make sense to refer to a class as "new", especially out of a context to a reference of what the "old" class was. Once this code is merged this class is current, not new.

tests/test_io_persistence.py
```diff
@@ -163,50 +163,18 @@ def test_butler_get_multiple_blend_parameter(self):
...
-            path=os.path.join(TESTDIR, "data", "v29_models.json"),
-            formatter="lsst.daf.butler.formatters.json.JsonFormatter",
-            refs=[ref],
-        )
-
-        # Ingest the legacy model into the butler
-        butler = makeTestCollection(repo, uniqueId="ingestion")
-        repo.registry.registerDatasetType(datasetType)
-        butler.ingest(dataset)
+        """A pre-``metadata`` (v29) archive loads and promotes its top-level
```
taranu: What does "top-level" add here?

tests/test_io_persistence.py
```diff
@@ -163,50 +163,18 @@ def test_butler_get_multiple_blend_parameter(self):
...
+        ``psf`` / ``psfShape`` into the typed ``model_psf`` attribute.
 
-        model = butler.get("old_scarlet_model_data", dataId={})
+        The ``(15, 15)`` model-PSF promotion is checked in
+        ``_load_legacy_model``; this is the regression test for finding IO-17
+        of ``audits/audit-2026-05-05.md``. Here we additionally pin that the
```
taranu: I don't see these audit docs in the repo so what use are references to them? And again this comment is overly verbose, containing details like the shape of the PSF and needless adjectives ("raw" key).

tests/test_io_persistence.py
```diff
@@ -163,50 +163,18 @@ def test_butler_get_multiple_blend_parameter(self):
...
-        # ``audits/audit-2026-05-05.md``.
-        self.assertIsNotNone(model.metadata)
-        self.assertIn("model_psf", model.metadata)
-        self.assertIsInstance(model.metadata["model_psf"], np.ndarray)
-        self.assertEqual(model.metadata["model_psf"].shape, (15, 15))
-        self.assertNotIn("psfShape", model.metadata)
-
-        test = butler.get("old_scarlet_model_data", dataId={}, parameters={"blend_id": 3495976385350991873})
-        self.assertEqual(len(test.blends), 1)
+        self.assertNotIn("psfShape", model.metadata or {})
```
taranu, line 172: Is this to keep the test passing if metadata becomes `None` in the future, assuming it's a dict now?

python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -21,82 +21,84 @@
...
+        The 2D model-space PSF shared by all bands.
+    psf
+        The per-band observed PSFs, shape ``(n_bands, height, width)``.
+    version
         The schema version of the serialized data.
     """
     model_type: str = MODEL_TYPE
+    blends: dict[int, scl.io.ScarletBlendBaseData]
     isolated: dict[int, IsolatedSourceData]
+    metadata: dict[str, Any] | None
```
taranu: I noticed this in the test but why would you want to let the metadata be `None` instead of at least an empty dict by default?

tests/test_io_persistence.py
```diff
@@ -274,14 +211,59 @@ def test_v30_legacy_model(self):
         self.assertEqual(float(iso.span_array.sum()), 119.0)
 
         # Single-blend parameter load also works on v30 archives.
+        self._assert_single_blend_load(butler, sorted(model.blends.keys())[0])
+
+    def test_v31a_legacy_model(self):
+        """A pre-DM-55109 archive promotes to the typed model on load.
+
+        ``data/v31a_models.json`` snapshots the on-disk layout written by the
```
taranu: Why is it named v31a when we're still on v30? Isn't this basically a 1.0.1 model?

tests/test_io_persistence.py
```diff
@@ -274,14 +211,59 @@ def test_v30_legacy_model(self):
...
+        release immediately before this ticket: model schema ``1.0.1`` with the
+        load-bearing model PSFs/bands in the metadata dict, and parent blends
```
taranu: There's that load-bearing again in another long and repetitive doc.

If the details of the schema changes are important enough then they should be in a changelog and/or a separate section of the docs and then the code can reference that instead of repeating the same explanations.

tests/test_io_persistence.py
```diff
@@ -397,6 +378,62 @@ def _test_blend(self, blendData1, blendData2, model_psf, psf, bands):
         )
         np.testing.assert_almost_equal(blend1.get_model().data, blend2.get_model().data)
 
+    def _load_legacy_model(self, filename, unique):
+        """Ingest a legacy JSON model fixture and return ``(model, butler)``.
```
taranu: Why are these being called fixtures anyway? It's a confusing choice in this context where you could be using pytest fixtures.

tests/test_io_hierarchical_blend_data.py
```diff
@@ -0,0 +1,138 @@
...
+from lsst.meas.extensions.scarlet.io import (
+    LsstHierarchicalBlendData,
+    LsstScarletModelData,
+    updateCatalogFootprints,
+)
+from lsst.scarlet.lite.io.blend_base import ScarletBlendBaseData
+from lsst.scarlet.lite.io.migration import MigrationRegistry
+
+
+def _spans(shape=(5, 6)):
```
taranu: I would call these `get_*` or `make_*`, but I suppose that's a matter of taste. I don't think there's any benefit to having single_under methods in a test module though.

tests/test_io_hierarchical_blend_data.py
```diff
@@ -0,0 +1,138 @@
...
+from lsst.meas.extensions.scarlet.io import hierarchical_blend_data as hbd_module
+from lsst.meas.extensions.scarlet.io import (
+    LsstHierarchicalBlendData,
+    LsstScarletModelData,
+    updateCatalogFootprints,
+)
+from lsst.scarlet.lite.io.blend_base import ScarletBlendBaseData
+from lsst.scarlet.lite.io.migration import MigrationRegistry
+
+
```
taranu, line 42: I think it would be preferable to define `DEFAULT_SHAPE = (5, 6)` and `DEFAULT_ORIGIN = (10, 20)` and use them in the function definitions below, as well as in the asserts that call them without args.

taranu (commented):
As usual, a mix of comments, questions and suggestions, which I'll elaborate on and summarize in Jira. I didn't see any obvious code correctness issues.

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
fred3m: I don't think so. I do think that for posterity it is important to keep this but I can remove the rst syntax in place of more readable single backticks.

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
fred3m: Because in theory `ndarray.shape` does not have to be 2D, so I would either have to treat the return type as `tuple[int, ...]` which is more general than it needs to be or explicitly write out `return (self.span_array.shape[0], self.span_array.shape[1])`.

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
fred3m: I'll reword this docstring.

TallJimbo:
Was using Pydantic for model<->JSON considered here?

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
fred3m: This should be faster than a loop, since it's internally 4 c calls as opposed to more lines in a python loop. I benchmarked and this way is slightly faster and easier to read.

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: None of the other dataclasses in scarlet lite work this way and I didn't see any reason that this one should be a different convention. I think that I adopted it before I knew abut `default_factory` and always initialized mutable defaults as None, since eg. `metadata: dict[str, any] = {}` is invalid.

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: I can make this comment more brief. The idea is that by lifting the functional (load bearing) parameters from the metadata there is now a contract that that they are obligated to meet; instead of being hidden, not typed checked, or even required, as metadata key value pairs.

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: I don't think that I want to start policing metadata. I take your point, but the entire point of this ticket is that metadata is only information to inform the user, and should in no way be used in downstream code. I was using it incorrectly before and it is likely to be None for all uses in the science pipelines.

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: same response as above

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: This is me slightly abusing metadata for persistence to disk. I made a design choice that changes to the schema should have minimal affect on the functions in `utils` that load and save the data to disk, as those aren't version controlled and any changes to the persistence API is a breaking change. To avoid that I still use the metadata field for storage. If I didn't (in this class only) then the zip file would need new keywords anytime we add or change parameters in this class. That seemed more unstable then just repurposing metadata for storage and unpacking it as needed.

Ironically this is perhaps Claude being too brief. 😄

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: That's a good point. This is exactly the part that I don't want to change much at all.

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: Where would you have me put it? In this case I think that this is very important context because it explains how legacy data is handled. As your other comments have made clear, the class docs itself don't need this story, so the place where the actual conversion of legacy code takes place seems like the  ideal place for it.

But I can make it more succinct.

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: No. I think that the first version of `ScarletBlendData` did not have `blend_type` at all, since ther was ly one type. It was added later when the `HierarchicalBlend` was created, so we need to support its absence.

python/lsst/meas/extensions/scarlet/io/utils.py
fred3m: I don't think so. The entire point of the migration system in scarlet lite is that legacy data is migrated on initialization. This should never trip and is really just being overprotective.

python/lsst/meas/extensions/scarlet/io/utils.py
fred3m: Yeah. In DM-55150 (the ticket that follows this one) I actually deprecate `psfs` in `Observation` and replace it with `psf`, so it will be consistent after that ticket.

tests/test_io_persistence.py
fred3m: There are multiple legacy versions of blend now. Some of them have `metadata is None` and others have an empty dict. So this needs to handle both.

tests/test_io_persistence.py
fred3m: I was surprised that this isn't spelled out in the dev guide anywhere because it's such a common practice, but there is no policy on docstrings for tests. The vast majority of our codebase has docstrings for test classes but not individual tests, so I consider test docstrings as gravy and if you don't want to read them then just don't.

But this issue with ids from an audit came up in a DM when Nate and I were discussing DM-54841. The thing that I want to persist is the exact bug, including the design choice made to fix it, in the test as either a docstring or comment. We ultimately agreed to keep references to the bug fix ideas as opposed to just the ticket number because there are 50+ bugs fixed in that ticket and this makes it easy for posterity to figure out why a test was necessary. We also agreed not to commit the audit file, keeping in practice with what Tim has been doing. But the doc is easily searchable in that PR because removing it was done on a single commit, making it easier for both AI and humans in the future to trace its origin.

tests/test_io_persistence.py
fred3m: There is already a v30 model. The schemas have changed since we created v30, so I needed to create a new set of models that will be in v31 (I know of changes in DM-55150, hence the a). I could call it `v30b` if you like, but then when we cut v31, the actual v31 models would have the wrong version id.

This is a tricky naming issue. The trouble is that the schema for each data class are independent, and designed that way on purpose because migration of eg. BlendData shouldn't depend on whether or not there were changes to SourceData. So I need some sort of versioning scheme for the data files in tests to ensure that they are always persistable in the future.

tests/test_io_persistence.py
fred3m: I can make this more brief but as I said above, you're free to skip the test docstrings if you like, by convention they aren't required. But I'll make it more brief, keeping a reference to the audit issue (and adding the ticket number) so that the raison d'etre of the test is preserved.

tests/test_io_persistence.py
fred3m: A fixture is a general (software) engineering term used in testing: https://en.wikipedia.org/wiki/Test_fixture. pytest also borrows the term for a specific object, and I get the confusion, but this is not a crazy out of left-field use of the term, it is standard terminology. But I agree that `test context` would be better.

fred3m:
@TallJimbo 
> Was using Pydantic for model<->JSON considered here?

Not anymore, this is built on top of the `scarlet_lite` serialization and migration system that was put in place in 2023 (it used to be a single [io](https://github.com/lsst/scarlet_lite/commit/0746420dd7639916690259a5c1b69fc8428efb6a) module) with guidance by Nate. I don't think that pydantic has this system of chained migration, or serializes numpy arrays natively, so using pydantic isn't likely to have saved us much, although it might have thwarted some of my earlier misguided efforts that abused the use of metadata.

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
fred3m: I don't remember exactly but I *think* that this is a mypy thing. It doesn't know what type `data["legacy_spans"]` is, so I have to cast it.

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: I opted not to do this. As you noted in a different comment, the code might change and then the docs are out of date. The only thing important here is that we now have a ModelData object that is independent of the scarlet_lite persisted model data so that we can do whatever we need for LSST and it won't affect the main scarlet lite repo.

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: Ok, I chose to update the docstring to explain what this is doing and greatly simplified this comment

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: Better solution: I moved the extraction of bands, model_psf, and psf to the from_dict method, which is where it belongs anyway. The old placement was necessary in scarlet lite but with the new migration this is unnecessary now. Technically I could convert it to a data class but in case we decide to ad functionality later I'll keep it as an ordinary class with a simple `__init__`.

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: As noted above, I just updated the class docstring to explain this

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
taranu: I think I prefer the latter but not particularly strongly.

python/lsst/meas/extensions/scarlet/io/hierarchical_blend_data.py
taranu: Ok, I'd just drop the backticks and :mod:/:meth: in favour of full words if they're not going to be parsed.

python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -21,82 +21,85 @@
...
         metadata: dict[str, Any] | None = None,
+        bands: tuple[str, ...] | None = None,
+        model_psf: np.ndarray | None = None,
+        psf: np.ndarray | None = None,
     ):
-        super().__init__(blends=blends, metadata=metadata)
+        self.blends = blends if blends is not None else {}
+        self.metadata = metadata
         self.isolated = isolated if isolated is not None else {}
+        # Extract attributes from the metadata blob if they were not passed
```
taranu: But it's not doing that anymore...

python/lsst/meas/extensions/scarlet/io/model_data.py
```diff
@@ -105,16 +108,38 @@ def as_dict(self) -> dict[str, Any]:
         -------
         result : dict[str, Any]
             The object encoded as a JSON-compatible dictionary.
+            The mechanism for serializing to a zip file is outside of the
```
taranu, line 142: This seems more like a Notes entry since it's aimed at developers more than users.

python/lsst/meas/extensions/scarlet/io/model_data.py
taranu: Hmm. It seems like you will eventually have to consider promoting metadata keys to full kwargs/attributes if they become important enough, but I suppose you'll figure that out if/when the time comes.

python/lsst/meas/extensions/scarlet/io/model_data.py
taranu: A doc page on schema versions could be useful, especially if you can make something like a table summarizing where/how attributes like origin and spans are stored. I suppose it depends on what likely use cases are - is this going to be encountered by e.g. DP2 users with a post-v30 branch, or...?

tests/test_io_persistence.py
taranu: I suppose it's a moot point since the audit file is in a previous commit, but I will note on Jira that these policies should be added to the dev guide via an RFC (even if they're just recommendations).

tests/test_io_persistence.py
taranu: So the v30 json = 1.0.0 and v31a is 1.0.1? I was just confused (and still am a bit) why you needed to distinguish between pipeline and schema version numbers. Are you saying you might need two different 1.0.1 test datasets, or...?

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: I think for the `LsstScarletModelData` object only the plan is for all new fields to be stored in metadata in `as_dict` and unpacked in `from_dict`. This will keep the IO functions stable and not affected by future migrations. I don't think that there's any loss of functionality or type checking in this, since that's purely an internal API and we have full control of packing and unpacking the fields in the metadata.

python/lsst/meas/extensions/scarlet/io/model_data.py
fred3m: We can do our best to protect users against changes in the code by migrating data to the current pipelines version. Unfortunately there is no procedure for deprecating class and function signatures and breakages like this are unfortunate casualties of that. I think that having messaging for users to use the version of the science pipelines used to produce the data is going to important messaging in general for the project. But I also agree that we should be communicating this information to users in some sort of doc. I'll include it in my work to create docs and tutorials on deblending in the next few weeks before the eDP2 release.

tests/test_io_persistence.py
fred3m: It's because each class data object has an independent schema version. If you look at `scarlet_lite`, where the migration system is written, Source, Blend, Component, etc. all have their own separate schema versions. This was a conscious decision to make it easier to track changes to individual objects and allow each object to migrate on its own, as opposed to having a master function that applied changes to an entire schema. This makes the migration system more modular and customizable by end users (for example the way that this package defines new class data objects outside of `scarlet_lite` and migrates them independently).

python/lsst/meas/extensions/scarlet/io/model_data.py
TallJimbo: I think class and function signatures is most of what our deprecation policies _do_ cover (certainly better than our lackof-policy on on-disk data deprecation).  But there are also a lot of not-really-public-but-not-obviously private classes and functions that that policy probably shouldn't be extended to literally, and this seems like it's one of them.

Anyhow, if the implication is ever that new code can't read old data, that is something we do want to approach with extreme caution.  I don't think we've ever knowingly broken the ability to read data we've shipped in a public release, even going back to HSC PDR1.

fred3m: Right. This system protects and ensures that users can read any old scarlet models. What I mean is that our deprecation system can't protect includes things like (what Dan is referring to here): in the previous version of the code the `psf`, `bands`, and `origin` of a set of scarlet models was stored in `metadata` but are now first class attributes of the class. I don't know of any way to warn users that loading a key from a dict is soon to be a deprecated key, and is part of the reason to move into class attributes.

Similarly, we don't (can't) deprecate arguments to a class or function, which is what I was referring to. There are ways around it, for example see the changes in observation.py in https://github.com/lsst/scarlet_lite/pull/26/changes , but changing args and kwargs is more complicated and not covered in the dev guide.

TallJimbo: My interpretation of the dev guide was that you can and should write manual `warnings.warning` calls for argument deprecations, but I suppose there are no examples so it's ambiguous.  But that's all irrelevant here because we definitely do not have a way to deprecate dictionary keys.

fred3m: Right, and I definitely do that. DM-55150 goes out of its way to even ensure that even ordered, non-keyword, arguments will still work with a warning. For this ticket I thought about overwriting the dict as a custom collection that would warn if one of the deprecated keys was used but that seemed like overkill.

fred3m: I changed my mind. This breaks a new update to metadetect and is likely to break other code, so I think that I am going to create a `DeprecatedDict`. I just posted in `#dm-algorithms-pipelines` to see if people think that this is of sufficiently general use to put in `lsst.utils.deprecated` or just add it to `meas_extensions_scarlet`.

PR #133, opened 2026-07-08
DM-55413: Compress scarlet model archives by default

fred3m:
Default write_scarlet_model to ZIP_DEFLATED; reads stay back-compatible with uncompressed archives. Add round-trip tests.

### DIFF
--- a/tests/test_io_utils.py
+++ b/tests/test_io_utils.py
@@ -21,8 +21,11 @@
 
 """Tests for the helpers in ``lsst.meas.extensions.scarlet.io.utils``."""
 
+import io
+import json
 import unittest
 import warnings
+import zipfile
 
 import lsst.meas.extensions.scarlet as mes
 import lsst.scarlet.lite as scl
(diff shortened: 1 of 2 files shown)


tests/test_io_utils.py
```diff
@@ -449,6 +451,103 @@ def test_loadBlend_requires_mCoadd(self):
...
+        # archive held by ``buf``.
+        buf.seek(0)
+        with zipfile.ZipFile(buf, "r") as zf:
+            return {info.compress_type for info in zf.infolist()}
+
+    def _assert_models_equal(self, modelData1, modelData2):
+        # The blends and isolated sources survive the round-trip.
+        np.testing.assert_almost_equal(modelData2.model_psf, modelData1.model_psf)
+        np.testing.assert_almost_equal(modelData2.psf, modelData1.psf)
+        self.assertEqual(set(modelData2.blends.keys()), set(modelData1.blends.keys()))
```
timj: Should we be checking values as well?

timj (approved):
Looks fine. My one minor comment for verifying backwards compatibility would be to be storing a small test fixture in the repo that you can demonstrate you can read and that the contents matches what you expect. In this case it likely doesn't really matter because the zip automatically uncompresses or not without your intervention.

PR #134, opened 2026-07-20
DM-55521: update deconvolve task to work with lsst.images

TallJimbo:


python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
```diff
@@ -203,6 +210,23 @@ class DeconvolveExposureConfig(
...
+        "This option only directly affects connection storage classes and hence 'runQuantum'; the 'run' "
+        "method behavior is determined by which type is actually passed in.",
+        allowed={
+            "legacy": (
+                "Read a lsst.cell_coadds.MultipleCellCoadd and restore 'background' (if useCellCoadd) or "
+                "lsst.afw.image.Exposure (if not useCellCoadd), and write an lsst.afw.image.Exposure."
+            ),
+            "future": (
+                "Read a lsst.images.cells.CellCoadd via 'connections.coadd' and write an "
+                "lsst.images.MaskedImage.  The useCellCoadd option will be ignore."
```
taranu: ignored

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
```diff
@@ -266,8 +296,25 @@ def run(
...
         -------
         deconvolved : `pipeBase.Struct`
-            Deconvolved exposure
+            Deconvolved exposure (if an `lsst.afw.image.Exposure` is provided;
+            an `lsst.images.MaskedImage` if an `lsst.images.cells.CellCoadd`
+            is provided).
         """
+        futureInputImage = None
+        if isinstance(coadd, CellCoadd):
+            # For now we just convert the future CellCoadd into an Exposure for
```
taranu, line 305: This makes sense but is there a follow-up ticket for this as a TODO? I imagine it's on Fred's radar but I wouldn't want it to get lost even if it's not the highest priority improvement.

python/lsst/meas/extensions/scarlet/scarletDeblendTask.py
```diff
@@ -1094,12 +1091,14 @@ def deblend(
...
-        scarletModelData :
-            The persistable data model for the deblender.
-        objectParents :
-            The parent catalog with deconvolved parents added.
+        result : `lsst.pipe.base.Struct`
+            A struct with the following attributes (all
+            `lsst.afw.table.SourceCatalog` instances):
+
+            - ``deblendedCatalog``: isolated and deblended child sources.
+            - ``scarletModelData``: persistable data model for the deblender.
```
taranu: I think "persistable deblended model data" would be preferable since "data model" sounds like a design for a format rather than the data itself.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
```diff
@@ -203,6 +210,23 @@ class DeconvolveExposureConfig(
         doc="Use cell-based coadd instead of regular coadd?",
         default=False,
     )
+    imageType = pexConfig.ChoiceField[str](
```
taranu, line 213: I suppose this field needing to be here and in `pipe_tasks` makes it more awkward to share even if you wanted to. `images` and `pipe_base` would be the only sensible options, no?

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
TallJimbo: No, and I'm not sure Fred actually does have this on his radar.  But I've sort of lost faith in Jira as a good place to remember big, long-term to-do items like this; it's really hard to scope the work into tickets this early, and then the tickets are too easy to lose.  And I also don't really want to push `lsst.images` conversions deep into the algorithmic code for its own sake, at least not now; I want it to be an option as a way to solve problems (and I do absolutely have problems I want to solve with it that will involve pushing it down into some tasks).  But in this case I'd rather leave it to Fred as something to work into the rest of his mental roadmap, and just let this comment be about how that might look.

python/lsst/meas/extensions/scarlet/deconvolveExposureTask.py
TallJimbo: And `images` is probably out because it doesn't want to depend on `pex_config`.  But I think the fact that the doc content is not the same rules out shared anyway.

